PR #1, opened 2025-10-09
Use the `bitcoin_p2p` crate

rustaceanrob:
Work in progress until there is an official `bitcoin = 0.33.0`, which may take many months.

First commit does the `bitcoin` update, second one swaps out the DNS resolution, third one updates the `BitcoinPeer` state machine.

### DIFF
--- a/src/peer.rs
+++ b/src/peer.rs
@@ -1,29 +1,25 @@
 use std::{
     collections::{HashMap, HashSet},
     fmt,
-    io::{Read, Write},
-    net::{SocketAddr, TcpStream},
-    ops::Deref,
+    net::SocketAddr,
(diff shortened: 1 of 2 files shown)


src/peer.rs
```diff
@@ -72,122 +75,52 @@ impl NodeState {
 ///       │ │
 ///       └─┘
 ///      Block
+#[derive(Default)]
 pub enum PeerStateMachine {
-    StartConnection,
-    Handshake(Handshake),
+    #[default]
```
TheCharlatan: Can you update the state diagram above?

TheCharlatan (approved):
lgtm

PR #11, opened 2025-10-29
Move node logic into dedicated binary

rustaceanrob:
This formalizes the node as a separate process and allows for new binaries, i.e. a client to stay in `bin`. I took the liberty of abreviating `kernel-node` as `node`.


TheCharlatan (approved):
lgtm

PR #12, opened 2025-10-29
Add `capnp` dependencies and build script

rustaceanrob:
Here we add a simple echo server and build the auto-generated code in `lib.rs`.


TheCharlatan (approved):
lgtm, seems to be generating correctly.

PR #14, opened 2025-10-30
Implement server logic for IPC interface

rustaceanrob:
`capnp-rpc` is dependent on the `tokio` runtime for running the RPC network. The `run` call will feature entirely blocking I/O, and `tokio` has a closure to handle these situations efficiently. I don't see any performance degregation on signet.

The `RpcNetwork` is not `Send`, so it has to be run on a `LocalSet` (basically a single-thread task manager). I have it bind to `./node.sock` with no configuration option, but maybe this should be tied to the data dir in the future. I have a dummy `IpcInterface` handle the `echo` requests, but you can imagine this having a chain state, wallet, etc in the future.

### DIFF
--- a/src/bin/node.rs
+++ b/src/bin/node.rs
@@ -15,13 +15,19 @@ use bitcoinkernel::{
     ChainType, ChainstateManager, ChainstateManagerOptions, Context, ContextBuilder, Log, Logger,
     SynchronizationState, ValidationMode,
 };
-use kernel_node::kernel_util::{ChainExt, DirnameExt};
 use kernel_node::peer::{BitcoinPeer, NodeState, TipState};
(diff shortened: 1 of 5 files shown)


src/bin/node.rs
```diff
@@ -299,7 +305,8 @@ fn run(
     Ok(())
 }
 
-fn main() {
+#[tokio::main]
+async fn main() {
```
TheCharlatan: Can we spawn the tokio runtime outside of the main thread so we can get clear separation between whatever runtime the kernel uses internally and to power its interfaces, and the tokio runtime required by capnp? The way I am reading this now, we're going full tokio on the entire runtime, but do we really want to commit to that?

PR #16, opened 2025-10-30
Implement IPC client logic

rustaceanrob:
AFICT the `configure_me` crate is not quite as ergonomic as `clap` when it comes to a program that will have many subcommands, so I re-added it here. For now the only client that is bootstrapped is the `echo` client, but this can easily be changed in the future.

~FYI I haven't actually tested this works yet, will do after lunch~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,6 +27,56 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


rustaceanrob:
In removing the global runtime from `main`, I forgot that the `LocalSet` still needs a runtime to execute. `Runtime::spawn` does not work as the `LocalSet` future is not `send`. Only option is to use `block_on`, but of course we don't want the execution thread to block, so I sent it off on a detached thread.

Echo interface works locally for me as of 9a3458ebcf24001f040d04acf10a4124a83158e9

TheCharlatan (approved):
That's the threading pattern I was expecting. Works on my end too.
LGTM

PR #24, opened 2025-12-29
Add block stale-ness check

rustaceanrob:
I noticed my home node got stuck on a block with the connection it had, but continued to exchange ping and pongs. This adds a very basic check to see if blocks are being withheld from us by disconnecting from a peer if they have not sent a block to us for 20 minutes.


sedited:
I've seen this too a couple of times. It doesn't really make sense to me. Seems fine to be rotating peers every now and again, but I wonder if at this point we should just introduce multi-peer messages handling.

PR #29, opened 2026-02-28
Add end-to-end sync tests

pzafonte:
Verify that `BitcoinPeer` correctly syncs blocks over a live P2P connection with real consensus validation. Tests run against a local bitcoind in regtest mode.

Adds four `#[ignore]` integration tests in `tests/e2e.rs`:
`e2e_sync_regtest_blocks`: mines 150 blocks, asserts kernel-node syncs to height 150.
`e2e_sync_resume`: syncs 50 blocks, mines 50 more, reconnects, and asserts the node resumes from height 50 without re-downloading. Exercises chainstate persistence and the locator-based fork-point logic.
`e2e_disconnect_reconnect`: syncs partway, stops bitcoind, restarts it with the same datadir, and asserts sync completes to height 150.
`e2e_mine_after_sync`: syncs 100 blocks, mines 10 more, reconnects, and asserts the node picks up new blocks via the getblocks/inv path.

Helpers: 
`KernelNode` wraps Context, ChainstateManager, tip_state across reconnections
`sync_to()` drives the message loop to a target height or timeout
`BitcoindInstance.stop()` gracefully shuts down with datadir reuse.

Requires `bitcoind` and `bitcoin-cli` in PATH.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,25 @@
 version = 4
 
 [[package]]
-name = "aho-corasick"
-version = "1.1.4"
+name = "addr2line"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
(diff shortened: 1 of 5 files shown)


pzafonte:
Closing in favor of better approach outlined in #76

PR #31, opened 2026-03-05
peer: Sync blocks from multiple peers in parallel

pzafonte:
Initial block download fetches blocks from a single peer, so the sync runs at that peer's pace and stalls when it is slow or unresponsive.

This connect to several peers at once, eight by default similar to Core, and split the block download between them. A shared queue hands out batches of hashes and an in-flight set keeps two peers from requesting the same block. When a peer disconnects, the blocks it still owed return to the queue.

This PR is following Bitcoin Core's behavior in the following ways:

| This PR                                                      | Bitcoin Core                                                 | Behavior                                                     |
| ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ |
| [`DEFAULT_MAX_PEERS = 8`](/kernel-node/kernel-node/blob/79b41d0/src/peer_manager.rs#L28) | [`MAX_OUTBOUND_FULL_RELAY_CONNECTIONS`](https://github.com/bitcoin/bitcoin/blob/v31.1/src/net.h#L69) | How many peers blocks are downloaded from at once |
| [`DOWNLOAD_BATCH_SIZE = 16`](/kernel-node/kernel-node/blob/79b41d0/src/peer.rs#L35) | [`MAX_BLOCKS_IN_TRANSIT_PER_PEER`](https://github.com/bitcoin/bitcoin/blob/v31.1/src/net_processing.cpp#L130) | Cap on how many blocks one peer is asked for at a time thus bounding the work lost if it drops |
| [`DownloadState::in_flight`](/kernel-node/kernel-node/blob/79b41d0/src/peer.rs#L52) | [`mapBlocksInFlight`](https://github.com/bitcoin/bitcoin/blob/v31.1/src/net_processing.cpp#L954) | Each block is requested from exactly one peer so parallel peers never duplicate a download |
| [`requeue_unreceived`](/kernel-node/kernel-node/blob/79b41d0/src/peer.rs#L112-L121) | [`FinalizeNode`](https://github.com/bitcoin/bitcoin/blob/v31.1/src/net_processing.cpp#L1698-L1708) | A departing peer's outstanding blocks become fetch-able again rather than being stranded |

### DIFF
--- a/src/peer.rs
+++ b/src/peer.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{HashMap, HashSet, VecDeque},
     fmt,
     net::SocketAddr,
     sync::{mpsc, Arc, Mutex},
@@ -27,6 +27,71 @@ use crate::kernel_util::{bitcoin_block_to_kernel_block, bitcoin_header_to_kernel
 const PROTOCOL_VERSION: ProtocolVersion = ProtocolVersion::INVALID_CB_NO_BAN_VERSION;
 const MAX_LOCATOR_HASHES: usize = 101;
 
+const DOWNLOAD_BATCH_SIZE: usize = 16;
+
+// No-op if the queue is already populated; multiple peers race to call this.
+fn populate_download_queue(chainman: &ChainstateManager, queue: &Mutex<VecDeque<BlockHash>>) {
+    if !queue.lock().unwrap().is_empty() {
+        return;
+    }
+    let active_height = chainman.active_chain().height();
+    let best = match chainman.best_entry() {
+        Some(entry) => entry,
+        None => return,
+    };
+    let best_height = best.height();
+    if best_height <= active_height {
+        return;
+    }
+    let count = (best_height - active_height) as usize;
+    let mut hashes = Vec::with_capacity(count);
+    let mut current = best;
+    while current.height() > active_height {
+        let hash = BlockHash::from_byte_array(current.block_hash().to_bytes());
+        hashes.push(hash);
+        match current.prev() {
+            Some(prev) => current = prev,
+            None => break,
+        }
+    }
+    hashes.reverse();
+    // Re-check: another thread may have populated the queue while we walked the chain.
+    let mut q = queue.lock().unwrap();
+    if !q.is_empty() {
+        return;
+    }
+    info!(
+        "Built download queue with {} blocks (heights {} to {})",
+        hashes.len(),
+        active_height + 1,
+        best_height
+    );
+    *q = VecDeque::from(hashes);
+}
+
+// Always locks queue first, then in_flight, to avoid deadlock.
+fn pop_download_batch(
+    queue: &Mutex<VecDeque<BlockHash>>,
+    in_flight: &Mutex<HashSet<BlockHash>>,
+    batch_size: usize,
+) -> Vec<BlockHash> {
+    let mut q = queue.lock().unwrap();
+    let mut in_flight = in_flight.lock().unwrap();
+    let mut batch = Vec::with_capacity(batch_size);
+    while batch.len() < batch_size {
+        match q.pop_front() {
+            Some(hash) => {
+                if !in_flight.contains(&hash) {
+                    in_flight.insert(hash);
+                    batch.push(hash);
+                }
+            }
+            None => break,
+        }
+    }
+    batch
+}
+
 #[derive(Clone)]
 pub struct TipState {
     pub block_hash: bitcoin::BlockHash,
@@ -46,6 +111,8 @@ pub struct NodeState {
     pub tip_state: Arc<Mutex<TipState>>,
     pub context: Arc<Context>,
     pub chainman: Arc<ChainstateManager>,
+    pub in_flight_blocks: Mutex<HashSet<BlockHash>>,
+    pub download_queue: Mutex<VecDeque<BlockHash>>,
 }
 
 impl NodeState {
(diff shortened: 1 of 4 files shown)


src/peer_manager.rs
```diff
@@ -0,0 +1,161 @@
...
+    peer::{BitcoinPeer, NodeState},
+};
+
+const TABLE_WIDTH: usize = 16;
+const TABLE_SLOT: usize = 16;
+const MAX_BUCKETS: usize = 4;
+
+pub type AddrTable = addrman::Table<TABLE_WIDTH, TABLE_SLOT, MAX_BUCKETS>;
+
+const DEFAULT_MAX_PEERS: usize = 4;
```
rustaceanrob: Bitcoin Core uses an outbound of 8 peers, seems fine to also have that here?

src/peer_manager.rs
```diff
@@ -0,0 +1,161 @@
...
+
+pub struct PeerManager {
```
rustaceanrob, line 36: I have further review for this type, but I think it would be appropriate to branch this one off into its own PR, even if we start with a peer manager of one peer?

rustaceanrob (commented):
Left an initial opinion regarding `PeerManager`

src/peer_manager.rs
pzafonte: Split into #95, which adds PeerManager on its own with a single connection. This PR now only adds the download queue.

src/peer_manager.rs
pzafonte: Set it to 8, and updated the PR description.

src/peer.rs
```diff
@@ -46,12 +47,103 @@ impl Default for TipState {
...
+        if self.next == fork_point || self.buffer.contains_key(&self.next) {
+            return;
+        }
+        debug!(target: Category::NET, "Re-anchoring block buffer to {}", fork_point);
+        self.next = fork_point;
+    }
+    fn pop_batch(&mut self, batch_size: usize) -> Vec<BlockHash> {
+        let mut batch = Vec::with_capacity(batch_size);
+        while batch.len() < batch_size {
+            match self.queue.pop_front() {
```
rustaceanrob:
```rust
let Some(hash) = self.queue.pop_front() else { break };
if self.in_flight.insert(hash) {
    batch.push(hash)
}  
```

src/peer.rs
```diff
@@ -46,12 +47,103 @@ impl Default for TipState {
...
+    // parent against the active chain needs get_block_tree_entry,
+    // which always returns None in bitcoinkernel 0.2.0.
+    // btck_ValidationInterfaceUpdatedBlockTip would supply the fork point directly.
+    fn reanchor_to(&mut self, fork_point: BlockHash) {
+        if self.next == fork_point || self.buffer.contains_key(&self.next) {
+            return;
+        }
+        debug!(target: Category::NET, "Re-anchoring block buffer to {}", fork_point);
+        self.next = fork_point;
+    }
```
rustaceanrob, line 84: Space here please

src/peer.rs
```diff
@@ -46,12 +47,103 @@ impl Default for TipState {
...
+    fn seed_tip(&mut self, tip: BlockHash) {
+        self.next = tip;
+    }
+
+    // populate_download_queue calls reanchor_to because checking a
+    // parent against the active chain needs get_block_tree_entry,
+    // which always returns None in bitcoinkernel 0.2.0.
+    // btck_ValidationInterfaceUpdatedBlockTip would supply the fork point directly.
+    fn reanchor_to(&mut self, fork_point: BlockHash) {
+        if self.next == fork_point || self.buffer.contains_key(&self.next) {
```
rustaceanrob: I think the second condition might be dead code. `buffer` is always drained immediately

rustaceanrob (commented):
Looks like a good start. I'm doing an IBD on mainnet right now, and I'm seeing a kind of stuttering behavior where every once and a while (say 10k blocks), the processing will be stuck on a block for a small amount of time. It's hard to tell if that is held up on kernel or the network. Either way, this is a good baseline to experiment with.

On memory,`8 peers * 16 blocks * 4mb` is about a half GB of memory at maximum. This is a worst case as well, so something I might try is increasing that to something like 96 blocks to see if that meaningfully affects the throughput.

src/peer.rs
pzafonte: Done.

src/peer.rs
pzafonte: Done.

src/peer.rs
pzafonte: Done.

pzafonte:
Made the suggested changes, rebased, and I'll run different batch sizes and report the metrics.

rustaceanrob:
Needs rebase

pzafonte:
Rebased.

Previously we kept our own copy of the tip in `TipState`, filled from the block checked callback. That's now replaced by asking the kernel directly. This branch followed the same pattern with the `next` pointer, so we can remove it, along with `TipState`, `seed_tip`, and `reanchor_to`.

Because we ask the kernel now, the check has to happen on the block processing thread, which owns the chainstate. Peers just buffer what arrives. Got rid of the block channel because the buffer passes the blocks now.

Previously we flushed a peer's leftovers once it had received everything it asked for, because a competing branch never gets a parent onto the active chain and those blocks would sit there forever. With a shared buffer no single peer finishing tells us that anymore, so now we wait until no peer has work left and send the oldest block we hold. `populate_download_queue` also skips blocks already buffered or being fetched, otherwise we'd never get there.

Also updated the README.

README.md
```diff
@@ -6,10 +6,12 @@ network. It is meant to showcase the limited initial API of the kernel library.
 It is not meant to be particularly performant, or robust against misbehaving
 peers.
 
-For now, IBD is done from a single peer, selected from the DNS seed nodes. If the
-connection to this peer happens to fail for some reason, a new peer will be selected.
-A direct connection can also be selected from the command line. See `--help` for
-this.
+IBD is done from several peers at once, selected from the DNS seed nodes. The
```
rustaceanrob: Thanks for updating, but I think we should perhaps remove this paragraph and add it to a `doc/design.md` or something. I think most users will just want to tinker with the binaries so we should keep the readme brief

rustaceanrob:
I'll re-review tomorrow, thanks!

pzafonte:
Ok, working on a proper multi-peer integration test, but which needs `--connect` to take more than one address. I'll have those up soon.

rustaceanrob (approved):
Reviewed with both an LLM and myself a few times. The logic makes sense to me here. I will let you update the readme and any modifications you would like to add before merge

pzafonte:
Okay, I'm going to do a little bit more performance/stress testing of the post-rebase changes. I'll push up the documentation changes and anything else if need be.

pzafonte:
Just updated the docs. I will likely add backpressure changes as a follow up, but I'll probably keep finding improvements, so I think this is good for now.

rustaceanrob (approved):
Looks good. I also have some thoughts on timeouts/scoring peers, etc, so we should have this in to experiment on top of. Good baseline!

PR #47, opened 2026-04-17
silent payments: integrate SPDK scanning and wallet IPC

pzafonte:
Integrate BIP-352 silent payment scanning into kernel-node using the `silentpayments` crate from [SPDK](https://github.com/cygnet3/spdk).
- Add `crates/wallet` as a workspace member with a `silentpayments` submodule that wraps SPDK's cryptographic primitives
- Scan each accepted block for payments using kernel undo data to recover prevout scripts
- Track received UTXOs and balance in an in-memory wallet
- Expose `import_keys`, `balance`, and `history` over a dedicated Cap'n Proto Unix socket (`wallet.sock`)
- Handle headers-first block announcements

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -631,6 +631,7 @@ dependencies = [
  "configure_me_codegen",
  "env_logger",
  "futures",
+ "hex",
  "libc",
  "log",
  "secp256k1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -74,14 +74,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2c8d66485a3a2ea485c1913c4572ce0256067a5377ac8c75c4960e1cda98605f"
 dependencies = [
  "bitcoin-internals",
- "bitcoin_hashes",
+ "bitcoin_hashes 0.14.1",
 ]
+
+[[package]]
+name = "bech32"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d86b93f97252c47b41663388e6d155714a9d0c398b99f1005cbc5f978b29f445"
 
 [[package]]
 name = "bech32"
 version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32637268377fc7b10a8c6d51de3e7fba1ce5dd371a96e342b34e6078db558e7f"
+
+[[package]]
+name = "bimap"
+version = "0.6.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "230c5f1ca6a325a32553f8640d31ac9b49f2411e901e427570154868b46da4f7"
 
 [[package]]
 name = "bindgen"
@@ -110,12 +122,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1e499f9fc0407f50fe98af744ab44fa67d409f76b6772e1689ec8485eb0c0f66"
 dependencies = [
  "base58ck",
- "bech32",
+ "bech32 0.11.1",
  "bitcoin-internals",
  "bitcoin-io",
  "bitcoin-units",
- "bitcoin_hashes",
- "hex-conservative",
+ "bitcoin_hashes 0.14.1",
+ "hex-conservative 0.2.2",
  "hex_lit",
  "secp256k1",
 ]
@@ -160,12 +172,21 @@ dependencies = [
 
 [[package]]
 name = "bitcoin_hashes"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "446819536d8121575eeb7e89efdbadb3f055e87e4bb66c6679a6d5cc2f4b64fd"
+dependencies = [
+ "hex-conservative 0.1.2",
+]
+
+[[package]]
+name = "bitcoin_hashes"
 version = "0.14.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "26ec84b80c482df901772e931a9a681e26a1b9ee2302edeff23cb30328745c8b"
 dependencies = [
  "bitcoin-io",
- "hex-conservative",
+ "hex-conservative 0.2.2",
 ]
 
 [[package]]
@@ -526,6 +547,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc0fef456e4baa96da950455cd02c081ca953b141298e41db3fc7e36b1da849c"
 
 [[package]]
+name = "hex"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
+
+[[package]]
+name = "hex-conservative"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "212ab92002354b4819390025006c897e8140934349e8635c9b077f47b4dcbd20"
+
+[[package]]
 name = "hex-conservative"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -602,6 +635,7 @@ dependencies = [
  "log",
  "tokio",
  "tokio-util",
+ "wallet",
 ]
 
 [[package]]
@@ -819,7 +853,7 @@ version = "0.29.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9465315bc9d4566e1724f0fffcbcc446268cb522e60f9a27bcded6b19c108113"
 dependencies = [
- "bitcoin_hashes",
+ "bitcoin_hashes 0.14.1",
  "rand",
  "secp256k1-sys",
 ]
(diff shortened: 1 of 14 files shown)


src/silentpayments/mod.rs
```diff
@@ -0,0 +1,1091 @@
...
+    tagged_hash("BIP0352/Inputs", &data)
+}
+
+pub fn compute_shared_secret<C: Verification>(
+    secp: &Secp256k1<C>,
+    scan_key: &ScanKey,
+    input_hash: &[u8; 32],
+    a_sum: &PublicKey,
+) -> Result<PublicKey, bitcoin::secp256k1::Error> {
+    let input_hash_scalar = Scalar::from_be_bytes(*input_hash)
```
rustaceanrob: A reference to a fixed size slice is passed here and is immediately de-referenced. Just pass `input_hash: [u8; 32]`

src/silentpayments/mod.rs
```diff
@@ -0,0 +1,1091 @@
...
+    tweaked_a.mul_tweak(secp, &scan_scalar)
+}
+
+pub fn derive_output_pubkey<C: Verification>(
+    secp: &Secp256k1<C>,
+    shared_secret: &PublicKey,
+    spend_key: &SpendKey,
+    k: u32,
+) -> Result<XOnlyPublicKey, bitcoin::secp256k1::Error> {
+    let mut data = Vec::with_capacity(33 + 4);
```
rustaceanrob: Prefer a fixed size 37 byte slice here and copy the data in.

src/silentpayments/mod.rs
```diff
@@ -0,0 +1,1091 @@
...
+}
+
+/// Derive B_spend_m = B_spend + H_tag("BIP0352/Label", b_scan || m) * G.
+pub fn derive_labeled_spend_key<C: Verification>(
+    secp: &Secp256k1<C>,
+    scan_key: &ScanKey,
+    spend_key: &SpendKey,
+    label: u32,
+) -> Result<SpendKey, bitcoin::secp256k1::Error> {
+    let mut data = Vec::with_capacity(32 + 4);
```
rustaceanrob: Same comment as above, this time with a 36 byte slice.

src/silentpayments/mod.rs
```diff
@@ -0,0 +1,1091 @@
...
+        for (i, b) in hex_decode(txid_hex).iter().rev().enumerate() {
+            op[i] = *b;
+        }
+        op[32..].copy_from_slice(&vout.to_le_bytes());
+        op
+    }
+
+    #[test]
+    #[allow(clippy::type_complexity)]
+    fn bip352_official_receiving_vectors() {
```
rustaceanrob: I think the word "official" can be dropped here. There is only one reference BIP

src/silentpayments/mod.rs
```diff
@@ -0,0 +1,1091 @@
...
+    let tag_hash = sha256::Hash::hash(tag.as_bytes());
+    let mut engine = sha256::Hash::engine();
+    engine.input(tag_hash.as_ref());
+    engine.input(tag_hash.as_ref());
+    engine.input(msg);
+    sha256::Hash::from_engine(engine).to_byte_array()
+}
+
+pub fn compute_input_hash(smallest_outpoint: &[u8; 36], a_sum: &PublicKey) -> [u8; 32] {
+    let mut data = Vec::with_capacity(36 + 33);
```
rustaceanrob: Can be a fixed slice at 69 bytes.

src/silentpayments/mod.rs
```diff
@@ -0,0 +1,1091 @@
...
+
+impl ScanKey {
+    pub fn from_slice(data: &[u8]) -> Result<Self, bitcoin::secp256k1::Error> {
+        let secp = Secp256k1::new();
+        let secret = SecretKey::from_slice(data)?;
+        let public = PublicKey::from_secret_key(&secp, &secret);
+        Ok(Self { secret, public })
+    }
+
+    pub fn secret_key(&self) -> &SecretKey {
```
rustaceanrob: These are both copy types. You can return them by value as opposed to reference.

src/silentpayments/mod.rs
```diff
@@ -0,0 +1,1091 @@
...
+pub fn compute_input_hash(smallest_outpoint: &[u8; 36], a_sum: &PublicKey) -> [u8; 32] {
+    let mut data = Vec::with_capacity(36 + 33);
+    data.extend_from_slice(smallest_outpoint);
+    data.extend_from_slice(&a_sum.serialize());
+    tagged_hash("BIP0352/Inputs", &data)
+}
+
+pub fn compute_shared_secret<C: Verification>(
+    secp: &Secp256k1<C>,
+    scan_key: &ScanKey,
```
rustaceanrob: Similar to the above comment, we should pass copy types by value, so just `ScanKey` here.

src/silentpayments/mod.rs
```diff
@@ -0,0 +1,1091 @@
...
+        let dir = std::env::temp_dir();
+        let path = dir.join("sp_wallet_test.bin");
+        wallet.save_to_file(&path).unwrap();
+        let loaded = SilentPaymentWallet::load_from_file(&path).unwrap();
+        assert_eq!(loaded.scan_height, 100);
+        assert_eq!(loaded.balance(), 50_000);
+        std::fs::remove_file(&path).ok();
+    }
+
+    fn hex_decode(s: &str) -> Vec<u8> {
```
rustaceanrob: Please remove this function in favor of the `hex` crate re-exported from `bitcoin`.

src/silentpayments/mod.rs
```diff
@@ -0,0 +1,1091 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use bitcoin::hashes::hash160;
+    use bitcoin::secp256k1::{PublicKey, Scalar, Secp256k1, SecretKey, XOnlyPublicKey};
+    use std::io;
+
+    #[test]
+    fn tagged_hash_differs_by_tag() {
```
rustaceanrob: I don't think we need any of the tests up to `make_keypair` function, all of these are tested in `rust-bitcoin`.

rustaceanrob:
7bde2e0a727c34e9a453c3062ee77ad2da4cf6f5 includes a new file that isn't actually linked to the project. It should otherwise not compile. Let's start by separating the commits such that each commit compiles, and the `silentpayments` module is included in `lib` at each step. I will continue code review in the meantime.

src/silentpayments/mod.rs
```diff
@@ -0,0 +1,1091 @@
+use bitcoin::hashes::{sha256, HashEngine};
+use bitcoin::secp256k1::{PublicKey, Scalar, Secp256k1, SecretKey, Verification, XOnlyPublicKey};
+
+mod scanning;
```
rustaceanrob: Everything referencing `scanning` and `wallet` should be removed until the subsequent commit.

yancyribbens:
I notice there is also [bip 352](https://docs.rs/bip352/latest/bip352/index.html) on crates.io.  Any reason to prefer re-writing it instead of importing the existing crate, forking the existing crate, or forking and add a PR back to existing crate?  If the existing bip352 crate is too unusable, maybe make a new crate for BIP 352?

rustaceanrob:
> Any reason to prefer re-writing it instead of importing the existing crate, forking the existing crate, or forking and add a PR back to existing crate?

I would prefer to not have a dependency that is bitcoin-related that we don't directly manage the release cycle of or is not well-reviewed and widely used (i.e. BDK). Reviewing a copy-paste of that crate into here would be essentially the same amount of review. While I agree it's not great to have a complete re-write, I think we should go ahead and do it here, and once BDK has support for silent payments, we do a drop-in replacement.

yancyribbens:
> I would prefer to not have a dependency that is bitcoin-related that we don't directly manage the release cycle of or is not well-reviewed and widely used (i.e. BDK). Reviewing a copy-paste of that crate into here would be essentially the same amount of review. While I agree it's not great to have a complete re-write, I think we should go ahead and do it here, and once BDK has support for silent payments, we do a drop-in replacement.

I'm not arguing against any or those points.  However, I do think it would make sense for this to be contained within it's own crate.

pzafonte:
Addressed all the feedback thus far in 7e03e6d

Should we extract this into its own crate in this PR?

yancyribbens:
> Should we extract this into its own crate in this PR?

I think so.

rustaceanrob:
I recommend something like `crates/wallet/silentpayments.mod` so we can keep both what we want in `crates` and what makes up a `wallet` flexible. I'll code review the first commit now, but I think it should be created in the first commit.

yancyribbens:
There is also an entire "dev kit" for silent payments in Rust: https://github.com/cygnet3/spdk.  I noticed this is being added to to electrs which is a mature project.

rustaceanrob:
Oh interesting...

I trust both the lead dev and the electrs guys. We should probably look into this.

rustaceanrob:
I reviewed SPDK and have some thoughts about path forward. The examples look intuitive and easy to integrate. Although there are some drawbacks, in that the crate brings in quite a few dependencies, I have more faith in them (electrs) maintaining the crate than bringing in custom SP logic here.

@pzafonte  I think we should repurpose this PR to integrate SPDK, however there is a blocker before you can get started. I was overly optimistic on the timeline for `bitcoin = 0.33` and the new `rust-bitcoin` crates. Some of the crates here depend on what is essentially a random commit in `rust-bitcoin`. You'll have to give me until about end-of-week to revert or add a branch to those crates to use `bitcoin = 0.32.8`. Once that is ready, we can get started with integrating SPDK.

rustaceanrob:
#48 is merged so `bitcoin` is now compatible with SPDK

pzafonte:
> #48 is merged so `bitcoin` is now compatible with SPDK

Great! I'll get started on this soon.

pzafonte:
Rewrote using SPDK and rebased on #49 
Everything updated in 96408a7, a149e9a, 42910cc, e8ec66d

yancyribbens:
> Rewrote using SPDK and rebased on https://github.com/kernel-node/kernel-node/pull/49

Don't be afraid to open a new PR.  This one is pretty scratched up with review comments that no longer apply.

pzafonte:
Opened fresh PR in #50

PR #49, opened 2026-04-29
Bump MSRV to `1.85.0`

rustaceanrob:
Ubuntu `26.04` supports Rust `1.91` and above. `bdk_wallet` has an MSRV of `1.85.0`. Given that we will likely want to add a wallet here, an MSRV bump seems appropriate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "kernel-node"
 version = "0.1.0"
 edition = "2021"
-rust-version = "1.84.0"
+rust-version = "1.85.0"
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


yancyribbens:
> Given that we will likely want to add a wallet here, an MSRV bump seems appropriate.

I always felt like it was an issue that bitcoin-core node packages it's wallet in the same repo and code-base as the node.  Agree though that it is useful to add some interface for wallet code.

yancyribbens:
Although, I think adding a wallet maybe as a workspace member so that it compiles together would be smart.

PR #50, opened 2026-05-02
silent payments: integrate SPDK scanning and wallet IPC

pzafonte:
New PR using SPDK... Integrate BIP-352 silent payment scanning into kernel-node using the `silentpayments` crate from [SPDK](https://github.com/cygnet3/spdk).
- Add `crates/wallet` as a workspace member with a `silentpayments` submodule that wraps SPDK's cryptographic primitives
- Scan each accepted block for payments using kernel undo data to recover prevout scripts
- Track received UTXOs and balance in an in-memory wallet
- Expose `import_keys`, `balance`, and `history` over the existing `node.sock`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,7 @@
+[workspace]
+members = [".", "crates/wallet"]
+resolver = "2"
+
 [package]
 name = "kernel-node"
 version = "0.1.0"
@@ -15,6 +19,7 @@ configure_me = "0.4.0"
 clap = { version = "4", features = ["derive"] }
 log = "0.4"
 env_logger = "0.10"
+wallet = { path = "crates/wallet" }
 ## IPC required dependencies
 capnp = "0.25"
 capnp-rpc = "0.25"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ rust-version = "1.85.0"
 # bitcoinkernel = { path = "../rust-bitcoinkernel", version = "0.0.15" }
 bitcoinkernel = "0.2"
 addrman = { package = "bitcoin-address-book", version = "0.1.1" }
-bitcoin = "0.32.8"
+bitcoin = { version = "0.32.8", features = ["rand-std"] }
 p2p = { package = "bitcoin-p2p", git =  "https://github.com/2140-dev/bitcoin-p2p.git", rev = "5dc03375636a10487c7e50659416876c28f58a3b" }
 ## Log and configuration
 configure_me = "0.4.0"
--- a/capnp/wallet.capnp
+++ b/capnp/wallet.capnp
@@ -1,0 +1,8 @@
+@0xb5d3a2f1e8c47690;
+
+interface Wallet {
+    importKeys @0 (scanKey :Data, spendKey :Data) -> (ok :Bool, message :Text);
+    getBalance  @1 () -> (sats :UInt64, scanHeight :UInt32, utxoCount :UInt32);
+    getHistory  @2 () -> (entries :Text);
+    receive     @3 () -> (address :Text);
+}
--- a/crates/wallet/src/silentpayments/scanning.rs
+++ b/crates/wallet/src/silentpayments/scanning.rs
@@ -1,0 +1,153 @@
+use ::silentpayments::{
+    receiving::{Label, Receiver},
+    utils::receiving::{calculate_ecdh_shared_secret, calculate_tweak_data, get_pubkey_from_input},
+};
+use bitcoin::consensus::encode;
+use bitcoin::secp256k1::{self, Scalar, SecretKey, XOnlyPublicKey};
+use bitcoin::{OutPoint, Script, Transaction};
+use bitcoinkernel::prelude::{
+    BlockSpentOutputsExt, CoinExt, ScriptPubkeyExt, TransactionExt, TransactionSpentOutputsExt,
+    TxOutExt,
+};
+
+use crate::silentpayments::wallet::Coin;
+
+pub struct InputData {
+    pub script_sig: Vec<u8>,
+    pub witness: Vec<Vec<u8>>,
+    pub prevout_script: Vec<u8>,
+    pub txid: String,
+    pub vout: u32,
+}
+
+pub fn scan_transaction(
+    receiver: &Receiver,
+    b_scan: &SecretKey,
+    inputs: &[InputData],
+    tx: &Transaction,
+) -> Vec<(usize, Scalar, Option<Label>)> {
+    let mut input_pub_keys = Vec::new();
+    let mut outpoints = Vec::new();
+    for input in inputs {
+        if let Ok(Some(pk)) =
+            get_pubkey_from_input(&input.script_sig, &input.witness, &input.prevout_script)
+        {
+            input_pub_keys.push(pk);
+            outpoints.push((input.txid.clone(), input.vout));
+        }
+    }
+
+    if input_pub_keys.is_empty() {
+        return vec![];
+    }
+
+    // Silent payments always produce taproot outputs — skip transactions without any.
+    let taproot_outputs: Vec<(usize, XOnlyPublicKey)> = tx
+        .output
+        .iter()
+        .enumerate()
+        .filter_map(|(i, out)| {
+            if out.script_pubkey.is_p2tr() {
+                XOnlyPublicKey::from_slice(&out.script_pubkey.as_bytes()[2..])
+                    .ok()
+                    .map(|pk| (i, pk))
+            } else {
+                None
+            }
+        })
+        .collect();
+
+    if taproot_outputs.is_empty() {
+        return vec![];
+    }
+
+    let pubkey_refs: Vec<&secp256k1::PublicKey> = input_pub_keys.iter().collect();
+    let tweak_data = match calculate_tweak_data(&pubkey_refs, &outpoints) {
+        Ok(td) => td,
+        Err(_) => return vec![],
+    };
+    let shared_secret = calculate_ecdh_shared_secret(&tweak_data, b_scan);
+
+    let xonly_outputs: Vec<XOnlyPublicKey> = taproot_outputs.iter().map(|(_, pk)| *pk).collect();
+    let found = match receiver.scan_transaction(&shared_secret, xonly_outputs) {
+        Ok(f) => f,
+        Err(_) => return vec![],
+    };
+
+    let mut result = Vec::new();
+    for (label, pubkey_map) in found.iter() {
+        for (pk, tweak) in pubkey_map {
+            if let Some((idx, _)) = taproot_outputs.iter().find(|(_, o)| o == pk) {
+                result.push((*idx, *tweak, label.clone()));
+            }
+        }
+    }
+    result
+}
+
+pub(crate) fn scan_block_inner(
+    receiver: &Receiver,
+    b_scan: &SecretKey,
+    kernel_block: &bitcoinkernel::Block,
+    spent_outputs: &bitcoinkernel::BlockSpentOutputs,
+    block_height: u32,
+) -> Vec<(OutPoint, Coin)> {
+    let mut found: Vec<(OutPoint, Coin)> = Vec::new();
+
+    // Skip coinbase; spent_outputs[i] maps to kernel_block.transactions().skip(1)[i].
+    for (kernel_tx, tx_spent) in kernel_block
+        .transactions()
+        .skip(1)
+        .zip(spent_outputs.iter())
+    {
+        let has_p2tr = kernel_tx.outputs().any(|out| {
+            let bytes = out.script_pubkey().to_bytes();
+            Script::from_bytes(&bytes).is_p2tr()
+        });
+        if !has_p2tr {
+            continue;
+        }
+
+        let tx_bytes = kernel_tx
+            .consensus_encode()
+            .expect("kernel tx serialization");
+        let btc_tx: Transaction =
+            encode::deserialize(&tx_bytes).expect("kernel tx deserialization");
+
+        let mut inputs = Vec::with_capacity(btc_tx.input.len());
+        for (input_idx, btc_input) in btc_tx.input.iter().enumerate() {
+            let coin = tx_spent
+                .coin(input_idx)
+                .expect("input/spent-output count mismatch");
+            inputs.push(InputData {
+                script_sig: btc_input.script_sig.as_bytes().to_vec(),
+                witness: btc_input.witness.iter().map(|item| item.to_vec()).collect(),
+                prevout_script: coin.output().script_pubkey().to_bytes(),
+                txid: btc_input.previous_output.txid.to_string(),
+                vout: btc_input.previous_output.vout,
+            });
+        }
+
+        let txid = btc_tx.compute_txid();
+        for (output_index, tweak, label) in scan_transaction(receiver, b_scan, &inputs, &btc_tx) {
+            if let Some(out) = btc_tx.output.get(output_index) {
+                let outpoint = OutPoint {
+                    txid,
+                    vout: output_index as u32,
+                };
+                found.push((
+                    outpoint,
+                    Coin {
+                        value: out.value,
+                        script_pubkey: out.script_pubkey.clone(),
+                        tweak,
+                        label,
+                        block_height,
+                        spent_by: None,
+                    },
+                ));
+            }
+        }
+    }
+    found
+}
--- a/crates/wallet/src/silentpayments/wallet.rs
+++ b/crates/wallet/src/silentpayments/wallet.rs
@@ -1,0 +1,208 @@
+use std::{
+    collections::{HashMap, HashSet},
+    fmt,
+};
+
+use bitcoin::secp256k1::{PublicKey, Scalar, SecretKey};
+use bitcoin::{hashes::Hash, Amount, OutPoint, ScriptBuf, Txid};
+use bitcoinkernel::prelude::{TransactionExt, TxInExt, TxOutPointExt, TxidExt};
+
+use crate::silentpayments::scanning::scan_block_inner;
+use crate::silentpayments::{Label, Network, Receiver};
+
+#[derive(Debug)]
+pub struct SilentPaymentKeys {
+    pub receiver: Receiver,
+    pub scan_key: SecretKey,
+}
+
+#[derive(Debug, Clone)]
+pub struct SpentBy {
+    pub txid: Txid,
+    pub block_height: u32,
+}
+
+#[derive(Debug, Clone)]
+pub struct Coin {
+    pub value: Amount,
+    pub script_pubkey: ScriptBuf,
+    pub tweak: Scalar,
+    pub label: Option<Label>,
+    pub block_height: u32,
+    pub spent_by: Option<SpentBy>,
+}
+
+pub enum HistoryEntry {
+    Received {
+        outpoint: OutPoint,
+        value: Amount,
+        block_height: u32,
+    },
+    Spent {
+        outpoint: OutPoint,
+        value: Amount,
+        spent_at: u32,
+    },
+}
+
+impl fmt::Display for HistoryEntry {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            HistoryEntry::Received {
+                outpoint,
+                value,
+                block_height,
+            } => write!(
+                f,
+                "recv {}:{} {} sats at {}",
+                outpoint.txid,
+                outpoint.vout,
+                value.to_sat(),
+                block_height
+            ),
+            HistoryEntry::Spent {
+                outpoint,
+                value,
+                spent_at,
+            } => write!(
+                f,
+                "spent {}:{} {} sats at {}",
+                outpoint.txid,
+                outpoint.vout,
+                value.to_sat(),
+                spent_at
+            ),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct Wallet {
+    pub scan_height: u32,
+    pub keys: Option<SilentPaymentKeys>,
+    pub spend_key: Option<PublicKey>,
+    pub network: Network,
+    utxos: HashMap<OutPoint, Coin>,
+}
+
+impl Wallet {
+    pub fn new(network: Network) -> Self {
+        Self {
+            scan_height: 0,
+            keys: None,
+            spend_key: None,
+            network,
+            utxos: HashMap::new(),
+        }
+    }
+
+    pub fn scan_block(
+        &mut self,
+        kernel_block: &bitcoinkernel::Block,
+        spent_outputs: &bitcoinkernel::BlockSpentOutputs,
+        block_height: u32,
+    ) -> usize {
+        let Some(keys) = self.keys.as_ref() else {
+            return 0;
+        };
+        let receiver = keys.receiver.clone();
+        let scan_key = keys.scan_key;
+
+        let found = scan_block_inner(
+            &receiver,
+            &scan_key,
+            kernel_block,
+            spent_outputs,
+            block_height,
+        );
+        let count = found.len();
+
+        for (outpoint, coin) in found {
+            self.utxos.insert(outpoint, coin);
+        }
+
+        for kernel_tx in kernel_block.transactions() {
+            let spending_txid = Txid::from_byte_array(kernel_tx.txid().to_bytes());
+            for input in kernel_tx.inputs() {
+                let kop = input.outpoint();
+                let prev = OutPoint {
+                    txid: Txid::from_byte_array(kop.txid().to_bytes()),
+                    vout: kop.index(),
+                };
+                if let Some(utxo) = self.utxos.get_mut(&prev) {
+                    utxo.spent_by = Some(SpentBy {
+                        txid: spending_txid,
+                        block_height,
+                    });
+                }
+            }
+        }
+
+        self.scan_height = block_height;
+        count
+    }
+
+    pub fn process_disconnect(&mut self, kernel_block: &bitcoinkernel::Block) {
+        let block_txids: HashSet<Txid> = kernel_block
+            .transactions()
+            .map(|tx| Txid::from_byte_array(tx.txid().to_bytes()))
+            .collect();
+
+        self.utxos
+            .retain(|outpoint, _| !block_txids.contains(&outpoint.txid));
+
+        for utxo in self.utxos.values_mut() {
+            if let Some(ref s) = utxo.spent_by {
+                if block_txids.contains(&s.txid) {
+                    utxo.spent_by = None;
+                }
+            }
+        }
+    }
+
+    pub fn utxo_count(&self) -> usize {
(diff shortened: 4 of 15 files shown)


capnp/wallet.capnp
```diff
@@ -0,0 +1,7 @@
+@0xb5d3a2f1e8c47690;
+
+interface Wallet {
```
rustaceanrob, line 3: Rather than introducing a separate interface and sock file, we should have sub-clients [as is done in Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/master/src/ipc/capnp/init.capnp). This prevents us from having to bootstrap with different RPC interfaces in the CLI code.

src/bin/node.rs
```diff
@@ -44,6 +49,113 @@ const STALE_BLOCK_DURATION: Duration = Duration::from_secs(60 * 20);
 
 configure_me::include_config!();
 
+fn to_wallet_network(network: Network) -> WalletNetwork {
```
rustaceanrob: For compatibility use cases, prefer an extension trait [as is done here](https://github.com/kernel-node/kernel-node/blob/master/src/kernel_util.rs#L9). You may also adapt that trait to have a `wallet_network` method.

src/bin/node.rs
```diff
@@ -44,6 +49,113 @@ const STALE_BLOCK_DURATION: Duration = Duration::from_secs(60 * 20);
...
+    let scan_key = match *wallet_state.scan_key.lock().unwrap() {
+        Some(k) => k,
+        None => return,
+    };
+    let spend_key = match *wallet_state.spend_key.lock().unwrap() {
+        Some(k) => k,
+        None => return,
+    };
+
+    let receiver = match build_receiver(&scan_key, spend_key, wallet_state.network) {
```
rustaceanrob: Appears strange to re-build the receiver every time we scan a block. Can this be done once at startup and be held in `WalletState`?

src/bin/node.rs
```diff
@@ -44,6 +49,113 @@ const STALE_BLOCK_DURATION: Duration = Duration::from_secs(60 * 20);
...
+
+    let undo = match chainman.read_spent_outputs(&entry) {
+        Ok(u) => u,
+        Err(e) => {
+            warn!("read_spent_outputs failed at height {block_height}: {e}");
+            return;
+        }
+    };
+
+    let raw = match kernel_block.consensus_encode() {
```
rustaceanrob: These should be added to `kernel_util` similar to the [existing](https://github.com/kernel-node/kernel-node/blob/master/src/kernel_util.rs#L66) conversion functions

src/bin/node.rs
```diff
@@ -44,6 +49,113 @@ const STALE_BLOCK_DURATION: Duration = Duration::from_secs(60 * 20);
...
+
+    let receiver = match build_receiver(&scan_key, spend_key, wallet_state.network) {
+        Ok(r) => r,
+        Err(e) => {
+            warn!("build_receiver failed: {e}");
+            return;
+        }
+    };
+
+    let entry = chainman.active_chain().tip();
```
rustaceanrob: Why is it certain that the block we are scanning corresponds to the active tip? I believe this function may need to accept a `BlockTreeEntry` to assert we are reading the correct undo data.

src/peer.rs
```diff
@@ -230,6 +230,21 @@ pub fn process_message(
                     (PeerStateMachine::AwaitingInv, vec![])
                 }
             }
+            NetworkMessage::Headers(headers) if !headers.is_empty() => {
```
rustaceanrob: Seems completely unrelated. Let's remove this for now as blocks should be announced by `inv` with our current implementation.

src/bin/node.rs
```diff
@@ -123,22 +234,6 @@ fn setup_logging() {
     unsafe { GLOBAL_LOG_CALLBACK_HOLDER = Some(Logger::new(KernelLog {}).unwrap()) };
 }
 
-// fn setup_validation_interface(
```
rustaceanrob, line 126: Please minimize the unrelated formatting changes.

src/peer.rs
pzafonte: Didn't feel great about it. Took a second look at this, realizing I designed my local round trip test script poorly, ended up tripping over myself.

yancyribbens:
Thanks for adding this.  I'm surprised how much code there is still even with including SPDK.  Although, I guess that's due to adding supporting wallet code.  I'd also prefer that SPDK used rust-bitcoin-coin-selection since it seems more robust and has more supported selection algos, although my opinion is maybe a bit biased :P

src/bin/node.rs
pzafonte: `build_receiver` is now called once inside `import_keys` and the result is stored in `WalletState.receiver`

src/bin/node.rs
pzafonte: Moved scanning into `with_block_connected_validation`, so `read_spent_outputs` called with `BlockTreeEntry`should read undo data for that block rather than the current tip

crates/wallet/src/silentpayments/scanning.rs
```diff
@@ -0,0 +1,104 @@
...
+
+pub fn scan_block(
+    receiver: &Receiver,
+    b_scan: &SecretKey,
+    block: &Block,
+    tx_input_data: Vec<Vec<InputData>>,
+) -> Vec<(bitcoin::Txid, FoundPayment)> {
+    let mut all_found = Vec::new();
+    // Skip coinbase (index 0); tx_input_data[i] maps to block.txdata[i+1].
+    for (i, tx) in block.txdata.iter().skip(1).enumerate() {
```
rustaceanrob: We can combine these two vectors with a `zip`. It also wouldn't hurt to add an assertion that for the invariant you commented above. The node should crash if the length of inputs does not match the length of outputs.

src/bin/node.rs
```diff
@@ -44,17 +47,101 @@ const STALE_BLOCK_DURATION: Duration = Duration::from_secs(60 * 20);
 
 configure_me::include_config!();
 
+fn scan_kernel_block(
+    chainman: &ChainstateManager,
+    kernel_block: &bitcoinkernel::Block,
+    entry: &bitcoinkernel::BlockTreeEntry,
+    wallet_state: &WalletState,
+) {
+    let receiver = match wallet_state.receiver.lock().unwrap().clone() {
```
rustaceanrob: Seems like we can use `as_ref`

src/bin/node.rs
```diff
@@ -44,17 +47,101 @@ const STALE_BLOCK_DURATION: Duration = Duration::from_secs(60 * 20);
...
+        Some(r) => r,
+        None => return,
+    };
+    let scan_key = match *wallet_state.scan_key.lock().unwrap() {
```
rustaceanrob: I think we need to panic here if there is no scan key because there is a receiver above. Is there a case where we can have a receiver with no scan key? Because this is the use of the API, should we move the scan key into the receiver?

src/bin/node.rs
```diff
@@ -403,6 +503,9 @@ fn main() {
         return;
     }
 
+    let node_sock_file = data_dir.clone() + "/node.sock";
```
rustaceanrob: Why is this variable being moved and re-declared? The diff is noisy here for not much reason.

crates/wallet/src/silentpayments/scanning.rs
pzafonte: Done. Thanks.

src/bin/node.rs
pzafonte: Woops, that's a leftover from when I had two separate sockets.

src/bin/node.rs
pzafonte: I double checked, and I believe receiver always needs a scan key. Created a struct that bundles Receiver and scan_key together so they're always paired.

crates/wallet/src/silentpayments/scanning.rs
```diff
@@ -0,0 +1,103 @@
...
+        return vec![];
+    }
+
+    // Silent payments always produce taproot outputs — skip transactions without any.
+    let taproot_outputs: Vec<(usize, XOnlyPublicKey)> = tx
+        .output
+        .iter()
+        .enumerate()
+        .filter_map(|(i, out)| {
+            let spk = out.script_pubkey.as_bytes();
```
rustaceanrob: This looks similar to [this method](https://docs.rs/bitcoin/latest/bitcoin/struct.Script.html#method.is_p2tr), which is already provided by `bitcoin`

crates/wallet/src/silentpayments/wallet.rs
```diff
@@ -0,0 +1,207 @@
...
+            .collect();
+
+        entries.sort_by_key(|e| match e {
+            HistoryEntry::Received { block_height, .. } => *block_height,
+            HistoryEntry::Spent { spent_at, .. } => *spent_at,
+        });
+        entries
+    }
+
+    pub fn save(&self, path: &Path) -> io::Result<()> {
```
rustaceanrob: There are likely more use cases for things that will be `save` and `load`. For instance a descriptor backup. I suggest doing the following:
- make an `io` module in `src`
- create a trait `FileExt`:
```rust
pub trait FileExt {
    save(&self, path: PathBuf) -> io::Result<()>;
    load(path: PathBuf) -> io::Result<Self>;
}
```
- implement it for the wallet

src/ipc.rs
```diff
@@ -35,4 +45,136 @@ impl server_capnp::server::Server for IpcInterface {
...
+        let text = history
+            .iter()
+            .map(|e| match e {
+                HistoryEntry::Received {
+                    outpoint,
+                    value,
+                    block_height,
+                } => {
+                    format!(
+                        "recv {}:{} {} sats at {}",
```
rustaceanrob: I think this would be cleaner if `HistoryEntry` implemented `Display`.

pzafonte:
Latest commits incorporate the most recent code reviews. Thanks!

yancyribbens:
@pzafonte , im curious what you think about using SPDK now that you've completed PRs both with and without.  Did you feel like the SPDK saved you a lot of work and effort vs the first PR you made?  Im somewhat skeptical because both PRs seem to be similar in size.

yancyribbens:
> Latest commits incorporate the most recent code reviews. Thanks!

btw, this is still in Draft.  Move it out of draft if you think it's a merge candidate.

pzafonte:
> @pzafonte , im curious what you think about using SPDK now that you've completed PRs both with and without. Did you feel like the SPDK saved you a lot of work and effort vs the first PR you made? Im somewhat skeptical because both PRs seem to be similar in size.

This was definitely much easier. W.R.T. size, it looks that way since I pushed up the spdk changes on the previous PR. Comparing just my original code (no test vectors), this PR is roughly ~1/3 the size. Much more maintainable, and we don't have to own the cryptography layer in this repo.

yancyribbens:
> and we don't have to own the cryptography layer in this repo.

However, I did notice there is a disclaimer we should keep in mind on the SPDK implementation:

> SPDK currently relies on cryptography that is not professionally reviewed. Be mindful of this when using SPDK with real funds.

rustaceanrob:
>  SPDK currently relies on cryptography that is not professionally reviewed. Be mindful of this when using SPDK with real funds

This will likely be delegated to `secp` once the C library has merged the API for silent payments.

src/bin/cli.rs
```diff
@@ -36,33 +39,53 @@ struct Echo {
...
-            capnp_rpc::rpc_twoparty_capnp::Side::Client,
-            Default::default(),
-        );
-        let mut rpc_system = capnp_rpc::RpcSystem::new(Box::new(network), None);
-        let client: server::Client =
-            rpc_system.bootstrap(capnp_rpc::rpc_twoparty_capnp::Side::Server);
-        tokio::task::spawn_local(rpc_system);
         match cli.commands {
             Commands::Echo(echo) => {
+                let client = connect_server(&datadir_path).await;
```
rustaceanrob: Can `connect_server` come before the `match`?

src/bin/cli.rs
```diff
@@ -36,33 +39,53 @@ struct Echo {
     message: String,
 }
 
+#[derive(Debug, Clone, clap::Subcommand)]
+enum WalletCmd {
+    /// Import a scan key (hex) and spend key (hex).
+    ImportKeys {
+        /// 32-byte scan private key as hex.
+        scan_key: String,
+        /// 33-byte spend public key as hex.
```
rustaceanrob: We need to expand on the docs here. This is the X-only public key right? Why do we need 33 bytes? If possible, we should import by X-only public key.

crates/wallet/src/silentpayments/wallet.rs
```diff
@@ -0,0 +1,243 @@
...
+
+        entries.sort_by_key(|e| match e {
+            HistoryEntry::Received { block_height, .. } => *block_height,
+            HistoryEntry::Spent { spent_at, .. } => *spent_at,
+        });
+        entries
+    }
+}
+
+impl FileExt for SilentPaymentWallet {
```
rustaceanrob: Can we add a test for roundtrips of this with dummy data? I would use `tempfile` as a dev dependency.

src/bin/node.rs
```diff
@@ -44,17 +47,98 @@ const STALE_BLOCK_DURATION: Duration = Duration::from_secs(60 * 20);
...
+        block_height
+    );
+}
+
 fn create_context(
     chain_type: ChainType,
     shutdown_tx: mpsc::Sender<()>,
     tip_state: &Arc<Mutex<TipState>>,
+    wallet_state: WalletState,
+    chainman_holder: Arc<std::sync::OnceLock<Arc<ChainstateManager>>>,
```
rustaceanrob, line 79: I had to remind myself of what `OnceLock` was. I'm not sure this is a great way to proceed. Have we tried using a `std::sync::mpsc::channel`? We can spawn a new thread that waits for blocks to scan when this callback is hit. So instead of `scan_kernel_block` here, we send the block over the channel to a consuming thread.

src/bin/node.rs
```diff
@@ -44,17 +47,98 @@ const STALE_BLOCK_DURATION: Duration = Duration::from_secs(60 * 20);
...
+
+    let undo = match chainman.read_spent_outputs(entry) {
+        Ok(u) => u,
+        Err(e) => {
+            warn!("read_spent_outputs failed at height {block_height}: {e}");
+            return;
+        }
+    };
+
+    let btc_block = kernel_block_to_bitcoin_block(kernel_block);
```
rustaceanrob: I think this is leaking some wallet logic to the `node` module, which we would like to keep as tidy as possible. It seems like we can refactor `scan_block` of the wallet to take `kernel::Block` and `kernel` spent outputs to move the entirety of the scanning logic to the wallet module.

src/bin/node.rs
```diff
@@ -44,17 +47,98 @@ const STALE_BLOCK_DURATION: Duration = Duration::from_secs(60 * 20);
...
+    wallet_state: WalletState,
+    chainman_holder: Arc<std::sync::OnceLock<Arc<ChainstateManager>>>,
 ) -> Arc<Context> {
     let shutdown_triggered = Arc::new(AtomicBool::new(false));
     let shutdown_triggered_clone = Arc::clone(&shutdown_triggered);
     let shutdown_tx_clone = shutdown_tx.clone();
     let tip_state_clone = tip_state.clone();
     Arc::new(ContextBuilder::new()
         .chain_type(chain_type)
+        .with_block_connected_validation(move |block: bitcoinkernel::Block, entry: bitcoinkernel::BlockTreeEntry<'_>| {
```
rustaceanrob, line 88: We also need to process [disconnected blocks](https://docs.rs/bitcoinkernel/latest/bitcoinkernel/state/context/struct.ContextBuilder.html#method.with_block_disconnected_validation) by removing coins that were created and spent in that block from our wallet history.

crates/wallet/src/silentpayments/wallet.rs
rustaceanrob: Would this make sense to go in the `io` module as well?

src/bin/cli.rs
pzafonte: I had it that way because `Receiver::new` takes 33-byte pubkey for spend_key, but we can import it as an X-only public key and then convert it.

src/bin/node.rs
pzafonte: I tried using the channel, and replaced `OnceLock` with a consumer thread that looks up the entry via `chainman.get_block_tree_entry(hash)` after the callback returns. A round trip test showed `get_block_tree_entry` returns None for just-connected blocks, even from a separate thread, so the entry isn't available at this stage. `read_spent_outputs(entry)` has to happen synchronously where the entry is alive. Since the callback is registered before `ChainstateManager` exists, we need some kind of slot for `chainman` or if the kernel callback delivered spent outputs alongside the block that might work, I think.  Open to alternatives if you can think of one, but `OnceLock` seems structurally required given the existing state of things.

rustaceanrob: As in we call `send` with the block and entry within `with_block_connected_validation` and the undo data is not present on the receiving thread? That seems strange if it is available within the callback.

pzafonte: Unless I'm missing something, to get the undo data you call `chainman.read_spent_outputs(entry)`, and the receiving thread doesn't have a reference to `chainman`.

rustaceanrob: The receiving thread can clone the `Arc` cheaply once it starts up. I will push up a branch as a demo for what I'm thinking tomorrow. Other review can be addressed in the meantime.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ log = "0.4"
 env_logger = "0.10"
 wallet = { path = "crates/wallet" }
 secp256k1 = "0.29"
+hex = "0.4"
```
rustaceanrob: This can be used from the `bitcoin` re-export

crates/wallet/src/silentpayments/wallet.rs
rustaceanrob: I just re-reviewed the rest of the PR and couldn't find where we actually call these. I realized as well that we would have to update this schema in the event we need additional information to spend, which I believe we do in the form of the input public key aggregate. Perhaps we entirely remove the I/O for now and decide how to proceed once sending is implemented? That would simplify this PR a bit.

src/bin/node.rs
rustaceanrob: So far I think you're right. Brutal. The self-referential nature of this architecture bothers me a lot but I think this is the only way to proceed. Confusing why the method returns `None` when it should return `Some`, at least at first glance.

rustaceanrob:
@pzafonte do you mind if we pair program the rest, i.e. I force-push this branch? You'll still get commit attribution, just eager to move this along while I have the time.

pzafonte:
> @pzafonte do you mind if we pair program the rest, i.e. I force-push this branch? You'll still get commit attribution, just eager to move this along while I have the time.

I don't mind that, I just pushed up the latest changes that address the existing review.

crates/wallet/src/silentpayments/wallet.rs
```diff
@@ -0,0 +1,180 @@
...
+use crate::silentpayments::scanning::kernel_block_to_bitcoin_block;
+
+#[derive(Debug, Clone)]
+pub struct SpentBy {
+    pub txid: Txid,
+    pub block_height: u32,
+}
+
+#[derive(Debug, Clone)]
+pub struct OwnedUtxo {
```
rustaceanrob: For each coin we will need the following:
- `ScriptPubKey`
- `OutPoint`
- A scalar representing the "tweak", should be returned at some point by `scan_transaction`
- `Amount`
- `Label` for change

src/ipc.rs
```diff
@@ -35,4 +45,109 @@ impl server_capnp::server::Server for IpcInterface {
...
+    ) -> Result<(), capnp::Error> {
+        let client: wallet_capnp::wallet::Client =
+            capnp_rpc::new_client(WalletIpcInterface::new(self.state.clone()));
+        results.get().set_wallet(client);
+        Ok(())
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct WalletState {
```
rustaceanrob: I think this should move into the commit that defines the wallet. The `ipc` module should only contain IPC related calls. The auto-generated code already contains a lot of boilerplate.

crates/wallet/src/silentpayments/wallet.rs
yancyribbens: Only the receiving side of silent payments is being added here, right?  Isn't this struct something that should be provided by the SPDK framework?

yancyribbens: fwiw, SPDK exposes the following for Receiver:

```
/// A struct representing a silent payment recipient.
///
/// It can be used to scan for transaction outputs belonging to us by using the [`scan_transaction`](Receiver::scan_transaction) function.
/// It optionally supports labels, which it manages internally.
/// Labels can be added with [`add_label`](Receiver::add_label).
#[derive(Debug, Clone, PartialEq)]
pub struct Receiver {
    version: u8,
    scan_pubkey: PublicKey,
    spend_pubkey: PublicKey,
    change_label: Label, // To be able to tell which label is the change
    labels: BiMap<Label, PublicKey>,
    pub network: Network,
}
```

crates/wallet/src/silentpayments/scanning.rs
```diff
@@ -0,0 +1,176 @@
...
+    utils::receiving::{calculate_ecdh_shared_secret, calculate_tweak_data, get_pubkey_from_input},
+};
+use bitcoin::consensus::encode;
+use bitcoin::{Amount, Block, ScriptBuf, Transaction, Txid};
+use bitcoinkernel::prelude::{
+    BlockSpentOutputsExt, CoinExt, ScriptPubkeyExt, TransactionSpentOutputsExt, TxOutExt,
+};
+use secp256k1::{Scalar, SecretKey, XOnlyPublicKey};
+
+pub struct InputData {
```
yancyribbens, line 15: Isn't this a `TxIn`? https://github.com/rust-bitcoin/rust-bitcoin/blob/0.32.x/bitcoin/src/blockdata/transaction.rs#L215

crates/wallet/src/silentpayments/scanning.rs
```diff
@@ -0,0 +1,176 @@
...
+
+pub struct FoundPayment {
+    pub output_index: usize,
+    pub pubkey: XOnlyPublicKey,
+    pub tweak: Scalar,
+    pub label: Option<Label>,
+}
+
+#[derive(Debug, Clone)]
+pub struct FoundUtxo {
```
yancyribbens: I believe this is `TxOut` with an added tweak and label: https://github.com/rust-bitcoin/rust-bitcoin/blob/0.32.x/bitcoin/src/blockdata/transaction.rs#L574

crates/wallet/src/silentpayments/scanning.rs
rustaceanrob: SPDK doesn't use `bitcoin` types so the conversion is unnecessary. That being said I think we can clean up the number of types introduced in this PR.

src/bin/cli.rs
```diff
@@ -81,6 +107,52 @@ fn main() {
...
             }
+            Commands::Wallet(cmd) => {
+                let wallet_response = client.make_wallet_request().send().promise.await.unwrap();
+                let client = wallet_response.get().unwrap().get_wallet().unwrap();
+                match cmd {
+                    WalletCmd::ImportKeys {
+                        scan_key,
+                        spend_key,
+                    } => {
+                        use bitcoin::hex::FromHex;
```
rustaceanrob: Imports at the top of the file please

rustaceanrob:
High level commit structure has good separation of concerns. I brainstormed a bit yesterday about final clean up for this PR if possible, here is my architectural feedback:

The distinction between `Wallet` and `WalletState` is not clear to me. I suggest we roll these into a single structure, `Wallet`. Within the `WalletState`, we have reference counted fields behind a mutex, but we should instead have a single `Arc` with a `Mutex` that locks the high level `Wallet` struct. This will minimize the heap allocations and make it clear at which points the wallet is being mutated.

It appears to me that we can also roll `scan_block` and `Wallet::process_block` into a single call. This makes the scope of where the wallet will be mutated to a single function call that we can evaluate. Having UTXOs be added after the block is "scanned" doesn't make much sense IMO.

We should probably try to avoid converting to a `bitcoin::Block` at all. We already process undo data directly from Core, it should be possible to check for taproot keys directly from the `kernel::Block` easily. Removing serialization roundtrips should be a top priority for us.

`OwnedUTXO` should just be called `Coin`.

We should also early-return in the callback functions of `ContextBuilder` if there is no wallet present.

src/bin/cli.rs
```diff
@@ -36,31 +40,54 @@ struct Echo {
     message: String,
 }
 
+#[derive(Debug, Clone, clap::Subcommand)]
+enum WalletCmd {
```
rustaceanrob, line 45: We probably want, with a disclaimer about private keys, a method to `GenerateKeys` that returns 1. scan private key 2. spend private key 3. spend x-only public key, all as hex. `OsRng` implements `CryptoRng`, which should be available in `rust-bitcoin` and can be enabled by feature if it isn't already. There is no need for an IPC call in this case, we can just generate and return in this file.

rustaceanrob:
Looking good. Does SPDK have tooling for generating the address? I'd like to give this a final smoke test by sending some signet coins.

yancyribbens:
> Does SPDK have tooling for generating the address?

https://github.com/cygnet3/spdk/blob/master/silentpayments/examples/create_wallet.rs

pzafonte:
> Looking good. Does SPDK have tooling for generating the address? I'd like to give this a final smoke test by sending some signet coins.

It has `silentpayments::SilentPaymentAddress::new()` I can write an example for deriving a silent payment address, and then write an end-to-end script to do the smoke test. Is there a better approach?

rustaceanrob:
I would like to develop a more thoughtful approach of how we do testing here, likely taking some inspiration from Floresta. In the meantime, let's add a final method `Receive` that returns the SP address string with no label.

rustaceanrob:
Epic. I tested the flow and was able to receive a payment on signet. I have some follow-ups in mind to clean this up and potentially make it more performant, but this is a good start.

ACK fac092194dd793ce7cd237a519e1110c5157eddd

sedited:
Nice! :D

pzafonte:
@rustaceanrob thanks for the review! as a follow up, I'm going to open a draft PR to add wallet persistence (I had already started on it, but kept it out for sake of scope) if you want to let me know if there's any overarching architectural preferences you have in mind.

rustaceanrob:
For next steps, I suggest implementing the sending half. This will indicate what fields are necessary to persist in case we missed anything the first time around. As an intermediate step, the node will also have to broadcast transactions, which I may work on.

yancyribbens:
FYI might be worth chiming in about adding more silent payment support upstream to rust-bitcoin https://github.com/rust-bitcoin/rust-bitcoin/discussions/5715

PR #53, opened 2026-05-18
Move silent payments block scanning onto separate thread

rustaceanrob:
The scanning process does CPU blocking work like reserialization of transactions and EC multiplications. This should occur on a separate thread so the validation callbacks don't block.

I was able to successfully receive a payment with this patch.

cc @pzafonte for review, as something similar was discussed in the original PR

### DIFF
--- a/crates/wallet/src/silentpayments/wallet.rs
+++ b/crates/wallet/src/silentpayments/wallet.rs
@@ -98,8 +98,8 @@ impl Wallet {
 
     pub fn scan_block(
         &mut self,
-        kernel_block: &bitcoinkernel::Block,
-        spent_outputs: &bitcoinkernel::BlockSpentOutputs,
+        kernel_block: bitcoinkernel::Block,
+        spent_outputs: bitcoinkernel::BlockSpentOutputs,
         block_height: u32,
     ) -> usize {
         let Some(keys) = self.keys.as_ref() else {
@@ -111,7 +111,7 @@ impl Wallet {
         let found = scan_block_inner(
             &receiver,
             &scan_key,
-            kernel_block,
+            &kernel_block,
             spent_outputs,
             block_height,
         );
@@ -142,7 +142,7 @@ impl Wallet {
         count
     }
 
-    pub fn process_disconnect(&mut self, kernel_block: &bitcoinkernel::Block) {
+    pub fn process_disconnect(&mut self, kernel_block: bitcoinkernel::Block) {
         let block_txids: HashSet<Txid> = kernel_block
             .transactions()
             .map(|tx| Txid::from_byte_array(tx.txid().to_bytes()))
--- a/src/bin/node.rs
+++ b/src/bin/node.rs
@@ -44,32 +44,37 @@ const STALE_BLOCK_DURATION: Duration = Duration::from_secs(60 * 20);
 
 configure_me::include_config!();
 
-fn scan_kernel_block(
-    chainman: &ChainstateManager,
-    kernel_block: &bitcoinkernel::Block,
-    entry: &bitcoinkernel::BlockTreeEntry,
-    wallet: &Arc<Mutex<Wallet>>,
-) {
-    let block_height = entry.height() as u32;
+enum ScanEvent {
+    Connected {
+        block_height: u32,
+        block: bitcoinkernel::Block,
+        spent_outputs: bitcoinkernel::BlockSpentOutputs,
+    },
+    Disconnected {
+        block: bitcoinkernel::Block,
+        block_height: u32,
+    },
+}
 
-    let spent_outputs = match chainman.read_spent_outputs(entry) {
-        Ok(u) => u,
-        Err(e) => {
-            warn!(target: Category::WALLET, "Reading spent outputs failed at height {block_height}: {e}");
-            return;
+#[derive(Clone)]
+struct FatalShutdown {
+    triggered: Arc<AtomicBool>,
+    tx: mpsc::Sender<()>,
+}
+
+impl FatalShutdown {
+    fn new(tx: mpsc::Sender<()>) -> Self {
+        Self {
+            triggered: Arc::new(AtomicBool::new(false)),
+            tx,
         }
-    };
+    }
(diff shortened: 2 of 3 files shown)


crates/wallet/src/silentpayments/wallet.rs
```diff
@@ -98,8 +98,8 @@ impl Wallet {
 
     pub fn scan_block(
         &mut self,
-        kernel_block: &bitcoinkernel::Block,
-        spent_outputs: &bitcoinkernel::BlockSpentOutputs,
+        kernel_block: bitcoinkernel::Block,
```
pzafonte, line 101: Is it intentional that `scan_block` takes `Block` by value, but `process_disconnect` takes `&Block`?

src/bin/node.rs
```diff
@@ -43,65 +43,56 @@ const STALE_BLOCK_DURATION: Duration = Duration::from_secs(60 * 20);
...
             let Some(chainman) = chainman_holder.get() else { return };
-            scan_kernel_block(chainman.as_ref(), &block, &entry, &wallet);
+            match chainman.read_spent_outputs(&entry) {
+                Ok(spent_outputs) => {
+                    if scan_tx.send(ScanEvent::Connected {
+                        block_height: entry.height() as u32,
+                        block,
+                        spent_outputs,
+                    }).is_err() {
+                        error!("Scan channel closed unexpectedly");
```
pzafonte: If the scan thread dies, every subsequent send() here errors and the wallet silently stops scanning. What's the intended recovery? The fatal/flush callbacks escalate to `shutdown_tx` should this do the same?

src/bin/node.rs
```diff
@@ -43,65 +43,56 @@ const STALE_BLOCK_DURATION: Duration = Duration::from_secs(60 * 20);
...
+                    }
+                }
+                Err(message) => {
+                    error!("Unable to read block spent outputs: {}", message);
+                }
+            }
         })
         .with_block_disconnected_validation(move |block: bitcoinkernel::Block, entry: bitcoinkernel::BlockTreeEntry<'_>| {
-            if wallet_for_disconnect.lock().unwrap().keys.is_none() {
```
pzafonte, line 98: Any way to skip `read_spent_outputs` during keyless IBD without re-coupling this callback to the wallet?

src/bin/node.rs
```diff
@@ -410,12 +441,15 @@ fn main() {
     let wallet = Arc::new(Mutex::new(Wallet::new(network.wallet_network())));
     let chainman_holder: Arc<std::sync::OnceLock<Arc<ChainstateManager>>> =
         Arc::new(std::sync::OnceLock::new());
+
+    let (scan_tx, scan_rx) = mpsc::channel::<ScanEvent>();
```
pzafonte, line 468: Is there a reason it's unbounded?  Could `sync_channel(N)` be a consideration here?

src/bin/node.rs
```diff
@@ -43,65 +43,56 @@ const STALE_BLOCK_DURATION: Duration = Duration::from_secs(60 * 20);
...
+                }
+            }
         })
         .with_block_disconnected_validation(move |block: bitcoinkernel::Block, entry: bitcoinkernel::BlockTreeEntry<'_>| {
-            if wallet_for_disconnect.lock().unwrap().keys.is_none() {
-                return;
+            if scan_tx_disconnect.send(ScanEvent::Disconnected {
+                block,
+                block_height: entry.height() as u32,
+            }).is_err() {
```
pzafonte, line 121: Same as comment on line 81

src/bin/node.rs
rustaceanrob: Makes sense for all of these to be escalated to fatal.

crates/wallet/src/silentpayments/wallet.rs
rustaceanrob: No, I'll update the ownership for disconnect block as well

src/bin/node.rs
rustaceanrob: If the channel becomes bounded we would have to implement some sort of back-pressure, which is possible but can start to get complicated. I also anticipate we could run into the same issue of undo or block data not being available on separate threads. For now I think it best to OOM, since the scanning thread should not be so CPU intensive as to work up a big queue, I doubt we would see an OOM in practice. We can revisit this if it becomes an issue.

crates/wallet/src/silentpayments/wallet.rs
rustaceanrob: Updated in bf698a09b91835f631c27b2476ebe942fc37c46e

src/bin/node.rs
rustaceanrob: In bf698a09b91835f631c27b2476ebe942fc37c46e I added a `FatalShutdown` struct to handle the duplicate logic between fatal flushes and now fatal wallet operations.

src/bin/node.rs
rustaceanrob: I thought about and I think passing the wallet is still the best approach. Otherwise we will have two different sources of truth for if there is a wallet present or not.  Re-implemented in c90043952d5be6a7be3ce81ae5ff48549c7081b0

rustaceanrob:
Received a payment on c90043952d5be6a7be3ce81ae5ff48549c7081b0 after addressing the feedback

src/bin/node.rs
rustaceanrob: Oh I see what you mean with the `sync_channel`, but still I don't think we should block in the callback hook if possible.

pzafonte:
Overall, It looks to be the case that the data stays available across threads by construction. I also ran some tests to sanity check, and no races.

src/bin/node.rs
pzafonte: > We can revisit this if it becomes an issue.

Agreed. Happy to defer.

rustaceanrob:
I'll wait for your ACK since you had part in this design

pzafonte:
ACK c900439

Issue #55, opened 2026-05-19
Silent payments tracking issue


rustaceanrob:
#### Features
- [x] Sending support and #51 #64 
- [x] Persistence #60 
- [x] File imports: JSON or binary format. Includes automatic file imports from configuration. #56 
- [ ] Offline signer support (i.e. BitBox)
- [x] Receiving #50 

#### Kernel API
- [x] Needs `TxIn`: `Witness`, `ScriptSig` https://github.com/bitcoin/bitcoin/pull/35380
- [x] Nice to have `ScriptPubkey::is_p2tr` https://github.com/sedited/rust-bitcoinkernel/pull/190

rustaceanrob:
@pzafonte I understand you've probably already started looking into some of these. If I had to suggest an ordering, I would focus on a file-based way to import/export the keys so we don't have to use terminal arguments, then a persistence layer, followed by sending once we have #51 implemented

pzafonte:
Yup, I looked into sending, file imports, and persistence. Happy to take on any of these of these tasks, including the Kernel API.

rustaceanrob:
Good ahead and put up any PRs related to these and I will tag them in this issue.

pzafonte:
@rustaceanrob Will revisit https://github.com/kernel-node/kernel-node/pull/50#pullrequestreview-4258392610 when we bump the version because it returns `Some` now!  https://github.com/sedited/rust-bitcoinkernel/pull/213

PR #56, opened 2026-05-22
silent payments: file-based key import/export

pzafonte:
Implements file imports from #55:

- `wallet::io::FileExt` trait that carries `save`/`load` so file-backed types can reuse the path handling.
- `wallet generate-keys --out <path>` writes a binary keys file (mode 0o600 on Unix, refuses to overwrite).
- `wallet import-keys-file <path>` reads it back.
- `--sp-keys-file <path>` (node option) auto-imports on startup; load errors warn and continue.

### DIFF
--- a/crates/wallet/src/silentpayments/keys_file.rs
+++ b/crates/wallet/src/silentpayments/keys_file.rs
@@ -1,0 +1,236 @@
+use std::{
+    fs,
+    io::{self, Write},
+    path::Path,
+};
+
+use bitcoin::secp256k1::{self, Secp256k1, SecretKey, XOnlyPublicKey};
+
+use crate::io::FileExt;
+
+const MAGIC: &[u8; 4] = b"SPKF";
+const KEY_LEN: usize = 32;
+const TAG_LEN: usize = 1;
+const FILE_LEN: usize = MAGIC.len() + KEY_LEN + TAG_LEN + KEY_LEN;
+
+const SPEND_TAG_SECRET: u8 = 0;
+const SPEND_TAG_XONLY_PUB: u8 = 1;
+
+/// `Secret` for software wallets. `XOnlyPublic` for hardware-signer / watch-only.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SpendKey {
+    Secret(SecretKey),
+    XOnlyPublic(XOnlyPublicKey),
+}
+
+impl SpendKey {
+    pub fn xonly(&self) -> XOnlyPublicKey {
+        match self {
+            SpendKey::Secret(sk) => {
+                let secp = Secp256k1::signing_only();
+                sk.public_key(&secp).x_only_public_key().0
+            }
+            SpendKey::XOnlyPublic(pk) => *pk,
+        }
+    }
+}
+
+/// On-disk format: 4-byte magic `SPKF` (Silent Payment Key File), 32-byte scan
+/// secret, 1-byte spend tag, 32-byte spend material. 69 bytes total.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SilentPaymentKeysFile {
+    pub scan_key: SecretKey,
+    pub spend: SpendKey,
+}
+
+#[derive(Debug)]
+pub enum KeysFileError {
+    Io(io::Error),
+    BadMagic,
+    WrongLength { expected: usize, actual: usize },
+    UnknownSpendTag(u8),
+    InvalidKey(secp256k1::Error),
+}
+
+impl std::fmt::Display for KeysFileError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            KeysFileError::Io(e) => write!(f, "i/o error: {e}"),
+            KeysFileError::BadMagic => write!(f, "bad magic: not a silent payments keys file"),
+            KeysFileError::WrongLength { expected, actual } => write!(
+                f,
+                "wrong file length: expected {expected} bytes, got {actual}"
+            ),
+            KeysFileError::UnknownSpendTag(t) => write!(f, "unknown spend tag: {t}"),
+            KeysFileError::InvalidKey(e) => write!(f, "invalid key: {e}"),
+        }
+    }
+}
+
+impl std::error::Error for KeysFileError {}
+
+impl From<io::Error> for KeysFileError {
+    fn from(e: io::Error) -> Self {
+        KeysFileError::Io(e)
+    }
+}
+
+impl From<secp256k1::Error> for KeysFileError {
+    fn from(e: secp256k1::Error) -> Self {
+        KeysFileError::InvalidKey(e)
+    }
+}
+
+impl SilentPaymentKeysFile {
+    pub fn new(scan_key: SecretKey, spend: SpendKey) -> Self {
+        Self { scan_key, spend }
+    }
+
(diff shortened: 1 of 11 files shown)


crates/wallet/src/silentpayments/keys_file.rs
```diff
@@ -0,0 +1,177 @@
...
+    fs,
+    io::{self, Write},
+    path::Path,
+};
+
+use bitcoin::secp256k1::{self, Secp256k1, SecretKey, XOnlyPublicKey};
+
+use crate::io::FileExt;
+
+const MAGIC: &[u8; 4] = b"KNSP";
```
rustaceanrob: What does the `KN` prefix stand for? Doesn't matter now, but I'd like to formalize our decisions with detailed docs at some point.

crates/wallet/src/silentpayments/keys_file.rs
```diff
@@ -0,0 +1,177 @@
...
+            spend_key,
+        })
+    }
+}
+
+impl FileExt for SilentPaymentKeysFile {
+    type Error = KeysFileError;
+
+    /// Refuses to overwrite (`AlreadyExists`). Mode 0o600 on Unix.
+    fn save(&self, path: &Path) -> Result<(), Self::Error> {
```
rustaceanrob, line 139: All of this can be simplified with [`File::create_new`](https://doc.rust-lang.org/std/fs/struct.File.html#method.create_new) which has the same semantics.

src/bin/cli.rs
```diff
@@ -71,6 +71,12 @@ enum WalletCmd {
         /// 32-byte x-only spend public key as hex.
         spend_key: String,
     },
+    /// Import silent payment keys from a binary file written by
+    /// `generate-keys --out`.
+    ImportKeysFile {
```
rustaceanrob: Because the CLI allows use even when the daemon isn't running, one can `generate-keys` then use the output on startup. I think we can drop this commit for now and just use the startup option.

src/bin/node.rs
```diff
@@ -444,6 +445,33 @@ fn run(
     Ok(())
 }
 
+fn auto_import_keys(wallet: &Mutex<Wallet>, path: &str) {
```
rustaceanrob: I know it is explicitly mentioned in the commit and PR that we log as an error, but I think this should be a crash. If we can't parse the file or it does not exist that is a clear bug and the user should attempt a recovery.

rustaceanrob: Also code nit, either the wallet should already be reference counted (`Arc<Mutex<`) or a `&mut Wallet` should be passed directly for function legibility.

rustaceanrob:
High level comment, we should allow the spend key to be passed by either private scalar or x-only public key in the file. This would allow us to use hardware signers and still use this import format. I suggest changing the file format to have an enum either `0` for private key or `1` for x-only key, followed by the bytes. Our default for `generate-keys` will be to write the private key.

crates/wallet/src/silentpayments/keys_file.rs
```diff
@@ -0,0 +1,177 @@
+#[cfg(unix)]
```
rustaceanrob: Can be removed when changing to `create_new`

yancyribbens: Also, since rust-bitcoin as already used as an import, why not try to use rust-bitcoin types such as `Magic`?

rustaceanrob: I think that would be an abuse of the type. `Magic` is used to identify the network. The scalars and public keys can apply to any network. We should use a new file identifier similar to PSBT as this is a new file type.

crates/wallet/src/silentpayments/keys_file.rs
pzafonte: Right. So my goal was to set permissions atomically, keeping the keys file readable only by the owner from the moment of creation which `File::create_new` doesn't do. Thinking about this a little more, perhaps a better way to approach this is restrict the process umask early in startup so new files default to owner-only permissions. Perhaps a follow up commit or PR.

crates/wallet/src/silentpayments/keys_file.rs
pzafonte: Kernel Node

rustaceanrob: Got it. Let's keep it more general as others might want to support it if it makes sense for them.

pzafonte: I see. How is SPKF  **S**ilent **P**ayments **K**ey **F**ile ?

rustaceanrob: Sounds good to me

crates/wallet/src/silentpayments/keys_file.rs
rustaceanrob: The deamonization logic has already burned Windows support, but I'd still like to be agnostic as much as we can with respect to operating systems, and I think `create_new` gets us 90 percent of the way there. We can follow up later if it makes sense to have more options for read and write permissions at startup.

src/bin/cli.rs
pzafonte: Done. Commit dropped.

src/bin/node.rs
pzafonte: Done. auto_import_keys now panics on either file-load failure or import failure, with a message identifying the offending path. A misconfigured --sp-keys-file aborts startup rather than logging a warning and continuing with an unimported wallet.

pzafonte: Done.

crates/wallet/src/silentpayments/keys_file.rs
pzafonte: Done. Now using `File::create_new()`

pzafonte:
> High level comment, we should allow the spend key to be passed by either private scalar or x-only public key in the file. This would allow us to use hardware signers and still use this import format. I suggest changing the file format to have an enum either `0` for private key or `1` for x-only key, followed by the bytes. Our default for `generate-keys` will be to write the private key.

keys_file.rs now has a SpendKey { Secret(SecretKey), XOnlyPublic(XOnlyPublicKey) } enum and a 1-byte tag.

rustaceanrob:
I'm going to pull and test locally but looking good to me

crates/wallet/src/silentpayments/wallet.rs
```diff
@@ -96,6 +96,18 @@ impl Wallet {
         }
     }
 
+    pub fn import_keys(
+        &mut self,
+        scan_key: SecretKey,
+        spend_xonly: XOnlyPublicKey,
+    ) -> Result<(), ::silentpayments::Error> {
+        let spend_pub = PublicKey::from_x_only_public_key(spend_xonly, Parity::Even);
```
rustaceanrob, line 104: I don't know much about the semantics of taproot, but I assume it's defined all x-only keys are even parity? I only ask because it seems strange the `from_x_only_public_key` function would need a parity argument at all if it's just going to be passed as `Even` every time

rustaceanrob:
Tested locally and received a payment. Nice job let's keep it rollin

ACK 6981b69429fa0a3a743c295b36287c347c3defc6

crates/wallet/src/silentpayments/wallet.rs
pzafonte: I believe ` from_x_only_public_key` is just the secp256k1 primitive and not taproot-specific.

PR #60, opened 2026-05-27
silent payments: file-backed wallet store

pzafonte:
Updated design:

`Wallet` gains `save` and `load` methods through `FileExt`. Each save writes the whole wallet to a binary file. To
avoid leaving a half-written file if the node crashes during a save, the save writes to a temporary file first and then renames it over the real one. `WalletStore` is type that holds the wallet file's path and the network the file should be for. It calls the wallet's save and load methods.

When the node starts, it loads the wallet. The scan thread saves the wallet through the store after every
block it processes. If a save fails, the node shuts down through `FatalShutdown` so it cannot keep scanning past data that never reached disk.

I have the `sp-wallet-store-backup` branch on my fork for reference.

### DIFF
--- a/crates/wallet/src/silentpayments/wallet_store.rs
+++ b/crates/wallet/src/silentpayments/wallet_store.rs
@@ -1,0 +1,608 @@
+use std::{
+    collections::HashMap,
+    fmt, fs,
+    io::{self, Write},
+    path::{Path, PathBuf},
+};
+
+use bitcoin::secp256k1::Scalar;
+use bitcoin::{hashes::Hash, Amount, OutPoint, ScriptBuf, Txid};
+
+use crate::io::FileExt;
+use crate::silentpayments::{Coin, Network, SpentBy, Wallet};
+
+// "S"ilent "P"ayments "W"allet "F"ile
+const MAGIC: &[u8; 4] = b"SPWF";
+const VERSION: u8 = 1;
+
+const HEADER_LEN: usize = 10;
+const COUNT_LEN: usize = 4;
+const RECORD_LEN: usize = 192;
+const SCRIPT_LEN: usize = 34;
+
+const NETWORK_MAINNET: u8 = 0;
+const NETWORK_TESTNET: u8 = 1;
+const NETWORK_REGTEST: u8 = 2;
+
+const SPENT_FLAG_UNSPENT: u8 = 0;
+const SPENT_FLAG_SPENT: u8 = 1;
+
+const LABEL_TAG_NONE: u8 = 0;
+const LABEL_TAG_SOME: u8 = 1;
+
+const R_SPENT_FLAG: usize = 0;
+const R_TXID: usize = R_SPENT_FLAG + 1;
+const R_VOUT: usize = R_TXID + 32;
+const R_VALUE: usize = R_VOUT + 4;
+const R_SCRIPT: usize = R_VALUE + 8;
+const R_TWEAK: usize = R_SCRIPT + SCRIPT_LEN;
+const R_LABEL_TAG: usize = R_TWEAK + 32;
+const R_LABEL_SCALAR: usize = R_LABEL_TAG + 1;
+const R_BLOCK_HEIGHT: usize = R_LABEL_SCALAR + 32;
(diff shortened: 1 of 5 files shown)


rustaceanrob:
Although we will probably want to move this way at some point in the future, given how simple our use case is, this might be overkill (thanks for demonstrating regardless). The way it integrates with the node and wallet appears awkward to me. I believe you had a PR that was more in line with what I'm about to recommend, so hopefully there's some code already present for these changes

1. I prefer we use our file extension trait and just implement `save` and `load` for `Wallet`. This means we will re-write the coins to the file on every save, but since this occurs on a separate thread and the file won't be big enough for this to have a significant impact until someone has hundreds of thousands of coins, this seems fine to me. It will greatly simplify the logic for our "database" and doesn't require changes to the scan and disonnect functions.
2. No new `WalletState` struct should have to be introduced. We can pass the file to the scanning thread and save each time there is a change using the extension trait. I can't imagine a user that would import keys and not want the result saved, so I think we should just store found coins on disk for all users. That should make the "integration" part nearly trivial. I'm fine with file I/O failures being fatal.

pzafonte:
Got it. I didn't have a clear sense of scale so I was making trade-offs that went beyond what we actually need. The separation between wallet code and file code, and the cost of rewriting the file, both matter less than I assumed. Easy enough to redo using the save/load pattern from the keys file PR.

pzafonte:
Pushed the rework. Updated the PR description to reflect the changes.

crates/wallet/src/silentpayments/wallet_store.rs
```diff
@@ -0,0 +1,608 @@
...
+use crate::io::FileExt;
+use crate::silentpayments::{Coin, Network, SpentBy, Wallet};
+
+// "S"ilent "P"ayments "W"allet "F"ile
+const MAGIC: &[u8; 4] = b"SPWF";
+const VERSION: u8 = 1;
+
+const HEADER_LEN: usize = 10;
+const COUNT_LEN: usize = 4;
+const RECORD_LEN: usize = 184;
```
rustaceanrob: We should pad this to 192 with dummy bytes for migration to additional fields if we need them and to make each record a cache-friendly size.

pzafonte: Updated in ee0a0a1

crates/wallet/src/silentpayments/wallet_store.rs
```diff
@@ -0,0 +1,608 @@
...
+
+const SPENT_FLAG_UNSPENT: u8 = 0;
+const SPENT_FLAG_SPENT: u8 = 1;
+
+const LABEL_TAG_NONE: u8 = 0;
+const LABEL_TAG_SOME: u8 = 1;
+
+const R_SPENT_FLAG: usize = 0;
+const R_TXID: usize = 1;
+const R_VOUT: usize = 33;
```
rustaceanrob: These would be more legible if they were `PREVIOUS_VAL + size of this` so `R_TXID + 4`

src/bin/node.rs
```diff
@@ -402,11 +409,19 @@ fn run(
...
                 }) => {
-                    let count =
-                        wallet
-                            .lock()
-                            .unwrap()
-                            .scan_block(block, spent_outputs, block_height);
+                    let mut wallet = wallet.lock().unwrap();
+                    let count = wallet.scan_block(block, spent_outputs, block_height);
+                    let now = (wallet.coin_count(), wallet.utxo_count());
+                    if now != last_saved {
```
rustaceanrob: Since we track the scan height anyway and the saves are atomic, let's write the file on every new block regardless of changes. I'm worried with logic like this that we somehow miss an update to the wallet and do not persist it. The naive approach at least buys us that we will never miss a wallet update. Same goes for `disconnect_block`. We can rehash this if a user reports this is a bottleneck.

rustaceanrob:
Is this ready for review?

pzafonte:
> Is this ready for review?

I believe so, the only major design I was stuck on was wallet updates and that's clarified now. I did some end to end tests locally and it appears to work.

rustaceanrob (approved):
Sent a payment and did a few startup/shutdowns. Works as expected.

PR #63, opened 2026-06-06
peer: support sendheaders block announcements 

pzafonte:
The node signals `sendheaders`, so peers announce new blocks with `headers` instead of `inv`. But the state machine only handles `headers` during initial block download, in `AwaitingHeaders`. Once synced, the connection sits in `AwaitingInv`, which matches only `inv`, so `headers` announcements fall through and get dropped, leaving the node stalled at its tip.

This handles `headers` in `AwaitingInv`. It validates the announced headers and requests the blocks directly with `getdata`, advancing to `AwaitingBlock`, or falls back to `getheaders` when the announcement does not connect to the node's header tree.

Reference: [BIP 130](https://bips.dev/130/)

### DIFF
--- a/src/peer.rs
+++ b/src/peer.rs
@@ -71,11 +71,12 @@ impl NodeState {
 /// ```text
 ///       [*]
 ///        │
-/// AwaitingHeaders
-///        ▼
-///   AwaitingInv
+/// AwaitingHeaders ◄──┐
+///        │           │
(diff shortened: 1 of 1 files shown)


rustaceanrob:
ACK bc0dfc8f57e7f3d16b17d718a5dbd222a639b9db

Issue #51, opened 2026-05-16
Broadcast transactions


rustaceanrob:
There is no transaction sharing logic within `node`. For the sending half of the wallet POC, a generalized broadcast queue has to be implemented.

yancyribbens:
Do you have an idea or framework in mind for implementing the send half?

rustaceanrob:
Sure, on a high level I think we will want to do something like a not-so-private broadcast. When a transaction needs to be sent, we:
1. spin up a connection to a new peer,
2. share the transaction, and 
3. wait for it to be advertised back by our current connection.

For an initial go at it, we can also omit 3.

For those that are less familiar, transactions are shared with a `wtxid`, then we wait for the peer to respond with `getdata`, then send the transaction. However, technically it is also possible to just send the transaction unsolicited.

This becomes more interesting with one-parent-one-child package relay but I will leave that as a follow up/feature request.

How this looks in Rust code I am agnostic to.

yancyribbens:
What about transaction building?  I thought that was part of the mentioned send half of wallet construction.

rustaceanrob:
That would fall under #55, which is also up for grabs. I don't have a strong opinion, but the BDK team has been putting a lot of time into `bdk-tx` so it is worth checking out.

yancyribbens:
> BDK team has been putting a lot of time into bdk-tx so it is worth checking out.

I was thinking of assembling something that uses the coin-selection lib I wrote.  That's one of the reasons I've started working on the P2P layer and following projects like this.  Although, I'm not sure that goal is very compatible with using BDK.

rustaceanrob:
#66 confirms this works

PR #64, opened 2026-06-10
silent payments: add sending support

pzafonte:
Adds a sendToAddress RPC and CLI command that take a silent payment address, an amount, and a fee rate, then build, sign, and broadcast the transaction. 

Notes:
- Coins chosen for a spend are reserved until the broadcast succeeds.
- Coin selection uses `bdk_coin_select`. It runs branch and bound to minimize the fee and falls back to a largest first pass when no solution is found. 
- Inputs and outputs are shuffled so the change output is not identifiable by position.

### DIFF
--- a/capnp/wallet.capnp
+++ b/capnp/wallet.capnp
@@ -6,4 +6,5 @@ interface Wallet {
     getHistory  @2 () -> (entries :Text);
     receive     @3 () -> (address :Text);
     broadcastRawTx @4 (tx :Data) -> (txid :Text);
+    sendToAddress @5 (address :Text, amountSat :UInt64, feeRateSatPerVb :Float64) -> (ok :Bool, message :Text);
 }
--- a/crates/wallet/src/silentpayments/sending.rs
+++ b/crates/wallet/src/silentpayments/sending.rs
@@ -1,0 +1,763 @@
+use std::collections::HashMap;
+use std::fmt;
+use std::str::FromStr;
+
+use bdk_coin_select::metrics::LowestFee;
+use bdk_coin_select::{
+    Candidate, ChangePolicy, CoinSelector, Drain, DrainWeights, Target, TargetFee, TargetOutputs,
+};
+use bitcoin::hashes::Hash;
+use bitcoin::key::TweakedPublicKey;
+use bitcoin::secp256k1::rand::seq::SliceRandom;
+use bitcoin::secp256k1::{self, Keypair, Message, Secp256k1, SecretKey, XOnlyPublicKey};
+use bitcoin::sighash::{Prevouts, SighashCache, TapSighashType};
+use bitcoin::transaction::Version;
+use bitcoin::{
+    absolute::LockTime, taproot, Address, Amount, FeeRate, OutPoint, ScriptBuf, Sequence,
+    Transaction, TxIn, TxOut, Weight, Witness,
+};
+use silentpayments::sending::generate_recipient_pubkeys;
+use silentpayments::utils::sending::calculate_partial_secret;
+use silentpayments::{Network, SilentPaymentAddress};
+
+use crate::silentpayments::wallet::{Coin, Wallet};
+
+const LONG_TERM_FEERATE_SAT_PER_VB: f32 = 1.0;
+
+#[derive(Debug)]
+pub enum SendError {
+    WatchOnly,
+    NoSpendableCoins,
+    DustAmount { amount: Amount, dust: Amount },
+    InsufficientFunds { needed: Amount, available: Amount },
+    NetworkMismatch,
+    InvalidRecipient(String),
+    OutputDerivation,
+    SilentPayments(::silentpayments::Error),
+    Secp(secp256k1::Error),
+    Sighash(bitcoin::sighash::TaprootError),
+}
+
+impl fmt::Display for SendError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            SendError::WatchOnly => {
+                write!(f, "wallet is watch-only: no spend secret to sign with")
+            }
+            SendError::NoSpendableCoins => write!(f, "no spendable coins"),
+            SendError::DustAmount { amount, dust } => write!(
+                f,
+                "amount {} sats is below the dust limit of {} sats",
+                amount.to_sat(),
+                dust.to_sat()
+            ),
+            SendError::InsufficientFunds { needed, available } => write!(
+                f,
+                "insufficient funds: need {} sats, have {} sats",
+                needed.to_sat(),
+                available.to_sat()
+            ),
+            SendError::NetworkMismatch => {
+                write!(f, "recipient address is for a different network")
+            }
+            SendError::InvalidRecipient(e) => write!(f, "invalid recipient address: {e}"),
+            SendError::OutputDerivation => write!(f, "recipient output key was not derived"),
+            SendError::SilentPayments(e) => write!(f, "silent payments error: {e}"),
+            SendError::Secp(e) => write!(f, "secp256k1 error: {e}"),
+            SendError::Sighash(e) => write!(f, "sighash error: {e}"),
+        }
+    }
+}
+
+impl std::error::Error for SendError {}
+
+impl SendError {
+    pub fn is_user_error(&self) -> bool {
+        matches!(
+            self,
+            SendError::WatchOnly
+                | SendError::NoSpendableCoins
+                | SendError::DustAmount { .. }
+                | SendError::InsufficientFunds { .. }
+                | SendError::NetworkMismatch
+                | SendError::InvalidRecipient(_)
+        )
+    }
+}
+
+impl From<::silentpayments::Error> for SendError {
+    fn from(e: ::silentpayments::Error) -> Self {
+        SendError::SilentPayments(e)
+    }
+}
+
+impl From<secp256k1::Error> for SendError {
+    fn from(e: secp256k1::Error) -> Self {
+        SendError::Secp(e)
+    }
+}
+
+impl From<bitcoin::sighash::TaprootError> for SendError {
+    fn from(e: bitcoin::sighash::TaprootError) -> Self {
+        SendError::Sighash(e)
+    }
+}
+
+pub enum Recipient {
+    SilentPayment(SilentPaymentAddress),
+    Address(Address),
+}
+
+impl Recipient {
+    pub fn parse(s: &str, network: Network) -> Result<Self, SendError> {
+        if let Ok(sp) = SilentPaymentAddress::try_from(s) {
+            if sp.get_network() != network {
+                return Err(SendError::NetworkMismatch);
+            }
+            return Ok(Recipient::SilentPayment(sp));
+        }
+        let address = Address::from_str(s)
+            .map_err(|e| SendError::InvalidRecipient(e.to_string()))?
+            .require_network(bitcoin_network(network))
+            .map_err(|_| SendError::NetworkMismatch)?;
+        Ok(Recipient::Address(address))
+    }
+}
+
+fn bitcoin_network(network: Network) -> bitcoin::Network {
+    match network {
+        Network::Mainnet => bitcoin::Network::Bitcoin,
+        Network::Testnet => bitcoin::Network::Testnet,
+        Network::Regtest => bitcoin::Network::Regtest,
+    }
+}
+
+struct SpendableCoin<'a> {
+    outpoint: OutPoint,
+    coin: &'a Coin,
+}
+
+impl Wallet {
+    pub fn build_transaction(
+        &self,
+        recipient: Recipient,
+        amount: Amount,
+        fee_rate: FeeRate,
+    ) -> Result<Transaction, SendError> {
+        let spend_secret = self.spend_secret.ok_or(SendError::WatchOnly)?;
+        let keys = self.keys.as_ref().ok_or(SendError::WatchOnly)?;
+        let change_address = keys.receiver.get_change_address();
+
+        let coins: Vec<SpendableCoin> = self
+            .utxos
+            .iter()
+            .filter(|(outpoint, coin)| coin.spent_by.is_none() && !self.reserved.contains(outpoint))
+            .map(|(outpoint, coin)| SpendableCoin {
+                outpoint: *outpoint,
+                coin,
+            })
+            .collect();
+
+        build_transaction(
+            &spend_secret,
+            recipient,
+            amount,
+            fee_rate,
+            self.scan_height,
+            change_address,
+            &coins,
+        )
+    }
+}
+
+fn build_transaction(
+    spend_secret: &SecretKey,
+    recipient: Recipient,
+    amount: Amount,
+    fee_rate: FeeRate,
+    tip_height: u32,
+    change_address: SilentPaymentAddress,
+    coins: &[SpendableCoin],
+) -> Result<Transaction, SendError> {
+    if coins.is_empty() {
+        return Err(SendError::NoSpendableCoins);
+    }
+    let secp = Secp256k1::signing_only();
+
+    let recipient_script = match &recipient {
+        Recipient::Address(address) => Some(address.script_pubkey()),
+        Recipient::SilentPayment(_) => None,
+    };
+    let recipient_dust = match &recipient_script {
+        Some(spk) => spk.minimal_non_dust(),
+        None => p2tr_dust(spend_secret, &secp),
+    };
+    if amount < recipient_dust {
+        return Err(SendError::DustAmount {
+            amount,
+            dust: recipient_dust,
+        });
+    }
+    let change_dust = p2tr_dust(spend_secret, &secp);
+    let recipient_weight = match &recipient_script {
+        Some(spk) => output_weight(spk.len()).to_wu(),
+        None => DrainWeights::TR_KEYSPEND.output_weight,
+    };
+
+    let feerate = cs_feerate(fee_rate);
+    let target = Target {
+        fee: TargetFee::from_feerate(feerate),
+        outputs: TargetOutputs::fund_outputs([(recipient_weight, amount.to_sat())]),
+    };
+    let change_policy = ChangePolicy::min_value_and_waste(
+        DrainWeights::TR_KEYSPEND,
+        change_dust.to_sat(),
+        feerate,
+        bdk_coin_select::FeeRate::from_sat_per_vb(LONG_TERM_FEERATE_SAT_PER_VB),
+    );
+
+    let (mut selected, drain) = select_coins(coins, target, change_policy, 100_000)?;
+    let mut rng = bitcoin::secp256k1::rand::thread_rng();
+    selected.shuffle(&mut rng);
+
+    let signing_keys: Vec<SecretKey> = selected
+        .iter()
+        .map(|c| spend_secret.add_tweak(&c.coin.tweak))
+        .collect::<Result<_, secp256k1::Error>>()?;
+
+    let change_value = drain.is_some().then(|| Amount::from_sat(drain.value));
+    let recipient_is_sp = matches!(recipient, Recipient::SilentPayment(_));
+    let derived: HashMap<SilentPaymentAddress, Vec<XOnlyPublicKey>> = if recipient_is_sp
+        || change_value.is_some()
+    {
+        // true marks each key as taproot since every silent payment coin is a P2TR output
+        let input_keys: Vec<(SecretKey, bool)> = signing_keys.iter().map(|k| (*k, true)).collect();
+        let outpoints: Vec<(String, u32)> = selected
+            .iter()
+            .map(|c| (c.outpoint.txid.to_string(), c.outpoint.vout))
+            .collect();
+        let partial_secret = calculate_partial_secret(&input_keys, &outpoints)?;
+        let mut sp_addrs = Vec::new();
+        if let Recipient::SilentPayment(sp) = &recipient {
+            sp_addrs.push(*sp);
+        }
+        if change_value.is_some() {
+            sp_addrs.push(change_address);
+        }
+        generate_recipient_pubkeys(sp_addrs, partial_secret)?
+    } else {
+        HashMap::new()
+    };
+
+    let recipient_script = match recipient {
+        Recipient::Address(address) => address.script_pubkey(),
+        Recipient::SilentPayment(sp) => sp_output_script(&derived, sp)?,
+    };
+    let mut output = vec![TxOut {
+        value: amount,
+        script_pubkey: recipient_script,
+    }];
+    if let Some(value) = change_value {
+        output.push(TxOut {
+            value,
+            script_pubkey: sp_output_script(&derived, change_address)?,
+        });
+    }
+    output.shuffle(&mut rng);
+
+    let input: Vec<TxIn> = selected
+        .iter()
+        .map(|c| TxIn {
+            previous_output: c.outpoint,
+            script_sig: ScriptBuf::new(),
+            sequence: Sequence::ENABLE_RBF_NO_LOCKTIME,
+            witness: Witness::new(),
+        })
+        .collect();
+
+    let mut tx = Transaction {
+        version: Version::TWO,
+        lock_time: LockTime::from_height(tip_height).unwrap_or(LockTime::ZERO),
+        input,
+        output,
+    };
+
+    let prevouts: Vec<TxOut> = selected
+        .iter()
+        .map(|c| TxOut {
+            value: c.coin.value,
+            script_pubkey: c.coin.script_pubkey.clone(),
+        })
+        .collect();
+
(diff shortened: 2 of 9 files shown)


yancyribbens:
> Coin selection uses bdk_coin_select. It runs branch and bound to minimize the fee and falls back to a largest first pass when no solution is found.

Any particular reason to use `bdk_coin_select` instead of `bitcoin_coin_select`?  FWIW `bitcoin_coin_select` has an interface [select_coins](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/blob/3825ab272829438ad18797792cabff106ce0d071/src/lib.rs#L117) which mirros that of bitcoin core.  That is, run Branch and Bound first, then, run Single Random Draw if BnB finds no solution.  Also, I maintain the 0.7.x branch which uses rust-bitcoin 0.32.x as a dependency to maintain compatibility.

pzafonte:
> Any particular reason to use `bdk_coin_select` instead of `bitcoin_coin_select`?

I was considering some beneficial aspects of `bdk-tx` and `bdk-sp` that might be useful to this project later on, but are not immediately useful in this PR, so there would be potential alignment in the future. Beyond that, I didn't do any comparison between coin selection libraries.

yancyribbens:
> I was considering some beneficial aspects of bdk-tx and bdk-sp that might be useful to this project later on, but are not immediately useful in this PR, so there would be potential alignment in the future. Beyond that, I didn't do any comparison between coin selection libraries.

I see.  Well I admit the `0.7.x` version is pretty hacky compared to more recent versions.  Although, I could try drafting a PR to use bitcoin-coin-selection at some point if you're open to it.  Also, there is the option to use coin-grinder as well with bitcoin-coin-selction which can do nifty optimizations dependent on fee environment for cost savings.

rustaceanrob:
Is there any reason to not upstream algos like coingrinder to `bdk_coin_select`? As much as possible I would like to use the most well reviewed and used crates in the ecosystem. When using BDK crates, we get all of the benefits when their users report bugs and add features to the crates at no cost to us, which is why I would opt to use `bdk_coin_select`. As Peter already mentioned as well, the integration with `bdk-tx` and hopefully `bdk-sp` will be tightly coupled.

yancyribbens:
> Is there any reason to not upstream algos like coingrinder to bdk_coin_select

Well, when I started work on the first coin-selection lib, there was no such thing as BDK.

> As much as possible I would like to use the most well reviewed and used crates in the ecosystem

I would argue that bitcoin-coin-selction in some ways is better reviewed than bdk-coin-select.  Murch has been kind enough to review my BnB, SRD and Coin-Grinder implementations over the years (and other areas as well).

> When using BDK crates, we get all of the benefits when their users report bugs and add features to the crates at no cost to us, which is why I would opt to use bdk_coin_select

For sure, BDK is a big project and popular framework (in  bitcoin land anyway), especially if you just want results fast.  I've worked for a few startups in the past that used BDK, and for them it made sense because they just wanted to get stuff off the ground, quick.

> As Peter already mentioned as well, the integration with bdk-tx and hopefully bdk-sp will be tightly coupled.

For sure.  I'm not super familiar with bdk-tx nor bdk-sp..  Understandably keeping within the bdk ecosystem may have some ease of use benefits.  Although, an open source project like this could use more specialized tooling than one might use if running a startup.

yancyribbens:
> Is there any reason to not upstream algos like coingrinder to bdk_coin_select

And I guess lastly I'd add that I upstream improvements back to bitcoin-core instead, which there have been a few here and there.

crates/wallet/src/silentpayments/sending.rs
```diff
@@ -0,0 +1,498 @@
...
+
+struct SpendableCoin<'a> {
+    outpoint: OutPoint,
+    coin: &'a Coin,
+}
+
+impl Wallet {
+    pub fn build_transaction(
+        &self,
+        recipient: SilentPaymentAddress,
```
rustaceanrob: Although we should definitely support sending to silent payments, I think the vast majority of recipients will be P2W*. We should parse any valid bitcoin address.

crates/wallet/src/silentpayments/sending.rs
```diff
@@ -0,0 +1,498 @@
...
+        fee: TargetFee::from_feerate(feerate),
+        outputs: TargetOutputs::fund_outputs([(
+            DrainWeights::TR_KEYSPEND.output_weight,
+            amount.to_sat(),
+        )]),
+    };
+    let change_policy = ChangePolicy::min_value(DrainWeights::TR_KEYSPEND, dust.to_sat());
+
+    let mut spendable: Vec<&SpendableCoin> = coins.iter().collect();
+    spendable.sort_by_key(|c| std::cmp::Reverse(c.coin.value));
```
rustaceanrob: Why would we sort these if they are just going into the `CoinSelector`?

yancyribbens: I think it's because he's using `select_until_target_met` later on if `bnb` fails.  That is why it's "largest first pass when no solution is found" presumably.

yancyribbens: I guess the sort could happen after bnb fails though so it doesn't get sorted twice, once by bnb and once here.

crates/wallet/src/silentpayments/sending.rs
pzafonte: Agreed. I will parse the recipient as a silent payment address first and then fall back to `bitcoin::Address` so any other valid address works.

crates/wallet/src/silentpayments/sending.rs
pzafonte: > I think it's because he's using `select_until_target_met` later on if `bnb` fails. That is why it's "largest first pass when no solution is found" presumably.

Yes, exactly that. The sort is only there for the largest first fallback. Good call on moving it.

yancyribbens: in O notation, it's the difference between O(n) and 2*O(n) I believe, so for performance its not a significant change.  Although, might help with readability :)

pzafonte:
Pushed an update, folded into the two commits:

- `build_transaction` now takes any standard bitcoin address.
- The largest first sort moved into the branch and bound fallback.
- `nLockTime` set to the chain tip for anti fee sniping

capnp/wallet.capnp
```diff
@@ -6,4 +6,5 @@ interface Wallet {
     getHistory  @2 () -> (entries :Text);
     receive     @3 () -> (address :Text);
     broadcastRawTx @4 (tx :Data) -> (txid :Text);
+    sendToAddress @5 (address :Text, amountSat :UInt64, feeRateSatPerVb :UInt64) -> (txid :Text);
```
rustaceanrob: `feeRateSatPerVb` should be a float given recent fee rates

crates/wallet/src/silentpayments/sending.rs
```diff
@@ -0,0 +1,694 @@
...
+            }
+        })?;
+    }
+    let drain = selector.drain(target, change_policy);
+    let selected = selector.apply_selection(&spendable).copied().collect();
+    Ok((selected, drain))
+}
+
+fn cs_feerate(fee_rate: FeeRate) -> bdk_coin_select::FeeRate {
+    bdk_coin_select::FeeRate::from_sat_per_wu(fee_rate.to_sat_per_kwu() as f32 / 1000.0)
```
rustaceanrob, line 355: What version of `bitcoin` are they on? Is this not the `bitcoin::FeeRate` type?

Not your fault, but this reads horrible and looks prone to a fee-high footgun.

yancyribbens: They are on `0.32`.  I've also noticed they have been slow to use Bitcoin types.  In someways that has been a frustration for me (using bitcoin types) since things like Amount don't perform well compared to native types.  I've spent a lot of time in optimizations so that the boundary points are Bitcoin types but internally sometimes revert to more primitive types when performance matters.  The other annoying thing is bdk select has no bench-marking or fuzz testing either.

yancyribbens: If I'm understanding this right, it looks like they use their own separate `FeeRate` (not a rust-bitcoin type): https://github.com/bitcoindevkit/coin-select/blob/master/src/feerate.rs

capnp/wallet.capnp
pzafonte: Done in the latest push. `feeRateSatPerVb` is now a `Float64`. I convert it to a `FeeRate` and round up so the effective rate is never below what was requested.

capnp/wallet.capnp
```diff
@@ -6,4 +6,5 @@ interface Wallet {
     getHistory  @2 () -> (entries :Text);
     receive     @3 () -> (address :Text);
     broadcastRawTx @4 (tx :Data) -> (txid :Text);
+    sendToAddress @5 (address :Text, amountSat :UInt64, feeRateSatPerVb :Float64) -> (txid :Text);
```
rustaceanrob: This could use a more generic message (i.e. ok/error variant with a string like "Successfully broadcast transaction `<txid>`

crates/wallet/src/silentpayments/sending.rs
```diff
@@ -0,0 +1,694 @@
...
+fn select_coins<'a, 'c>(
+    coins: &'c [SpendableCoin<'a>],
+    target: Target,
+    change_policy: ChangePolicy,
+    max_bnb_rounds: usize,
+) -> Result<(Vec<&'c SpendableCoin<'a>>, Drain), SendError> {
+    let spendable: Vec<&SpendableCoin> = coins.iter().collect();
+    let candidates: Vec<Candidate> = spendable
+        .iter()
+        .map(|c| Candidate::new_tr_keyspend(c.coin.value.to_sat()))
```
yancyribbens, line 327: From what I've read about Silent payments, the receiver is always a taproot output.  However, I believe the sender can use any output type to send, right?  Why then is every candidate output using `tr_keyspending` (assuming that tr is Tap Root here).  In effect, I think you are telling the selection algo that all inputs are taproot inputs, but I don't think this is necessarily correct.

crates/wallet/src/silentpayments/sending.rs
```diff
@@ -0,0 +1,694 @@
...
+    let drain = selector.drain(target, change_policy);
+    let selected = selector.apply_selection(&spendable).copied().collect();
+    Ok((selected, drain))
+}
+
+fn cs_feerate(fee_rate: FeeRate) -> bdk_coin_select::FeeRate {
+    bdk_coin_select::FeeRate::from_sat_per_wu(fee_rate.to_sat_per_kwu() as f32 / 1000.0)
+}
+
+fn output_weight(script_len: usize) -> u64 {
```
yancyribbens: For clarity and type safety, you might consider returning a Weight type here: https://github.com/rust-bitcoin/rust-bitcoin/blob/master/units/src/weight.rs.  A `u64` usually indicates you're returning something measured in vB (virtual bytes), although you are multiplying by 4, so I think this is actually a Weight Unit type.

crates/wallet/src/silentpayments/sending.rs
```diff
@@ -0,0 +1,694 @@
...
+    // 8-byte value, 1-byte length prefix, then the script, times 4 weight units per byte.
+    ((8 + 1 + script_len) * 4) as u64
```
yancyribbens: `as` _looks_ ok here, but generally it can be sketchy to use `as` since you may truncate bytes and the program would silently continue.

crates/wallet/src/silentpayments/sending.rs
pzafonte: That is true _in general_ but for this wallet the coins being spent here are exclusively silent payment outputs received by scanning, so every coin in this wallet is `new_tr_keyspend`, but if the wallet later holds other input types the candidate weight would need to vary per input.

yancyribbens: I see.  Good to know, that does make the spend case in this wallet easier to handle.

yancyribbens: I would also verify that this is default taproot sighash size vs non-default.  It's only a byte difference, but if the wallet is for some reason creating non-default, that could add up in the case where lots of inputs are used.

crates/wallet/src/silentpayments/sending.rs
```diff
@@ -0,0 +1,694 @@
...
+        Some(spk) => output_weight(spk.len()),
+        None => DrainWeights::TR_KEYSPEND.output_weight,
+    };
+
+    let feerate = cs_feerate(fee_rate);
+    let target = Target {
+        fee: TargetFee::from_feerate(feerate),
+        outputs: TargetOutputs::fund_outputs([(recipient_weight, amount.to_sat())]),
+    };
+    let change_policy = ChangePolicy::min_value(DrainWeights::TR_KEYSPEND, change_dust.to_sat());
```
yancyribbens: I'm confused by the use of `change_dust` here.  Generally, the `ChangePolicy` (see also `cost_of_change` in bitcoin core) is the cost to create a change output.  The reason this matters is that the bnb solution can be exceeded by at most `cost_of_change` before the solution needs to worry about what to do with the excess.  If a solution is found that is greater than the target and less than target + cost to produce a change output, then the solution is considered acceptable since there is not enough leftover to worry creating a change output.  If target + cost of change is exceeded, then a different method can be used, like SRD.  Instead, you are using `change_dust` and I'm wondering if this was intentional?

pzafonte: Got it, my intention was just to avoid creating dust change outputs but it looks like I can use `ChangePolicy::min_value_and_waste`

crates/wallet/src/silentpayments/sending.rs
pzafonte: Thanks. `Weight::from_wu_usize` will naturally resolve this.

crates/wallet/src/silentpayments/sending.rs
```diff
@@ -0,0 +1,694 @@
...
+    } else {
+        HashMap::new()
+    };
+
+    let recipient_script = match recipient {
+        Recipient::Address(address) => address.script_pubkey(),
+        Recipient::SilentPayment(sp) => sp_output_script(&derived, sp)?,
+    };
+    let mut output = vec![TxOut {
+        value: amount,
```
yancyribbens, line 257: If im not mistaken, you are creating a change output that is the size of the transaction target amount.  The function `build_transaction` takes a variable `amount`, and I don't see that amount be changed, then it's used to create a change output.  This should either be a changeless transaction in which case there is no change output, or the change should be the excess of what was found by the selection algorithm.

yancyribbens: s/be changed/is changed

capnp/wallet.capnp
pzafonte: Added the message, along with the ok flag and expected errors that are returned using it.

crates/wallet/src/silentpayments/sending.rs
pzafonte: That's the recipient output, not the change. The first output pays the recipient, so it's amount:
```
let mut output = vec![TxOut {
    value: amount,
    script_pubkey: recipient_script,
}];
```
The change is the next output, with value change_value:
```
if let Some(value) = change_value {
    output.push(TxOut {
        value,
        script_pubkey: sp_output_script(&derived, change_address)?,
    });
}
```
change_value is the drain value from the selector:
`let change_value = drain.is_some().then(|| Amount::from_sat(drain.value));`

which comes from selector.drain:
`let drain = selector.drain(target, change_policy);`

So when drain is None, change_value is None and no change output is added, which is the changeless case. When drain is Some, the change output value is drain.value, the leftover from selection, not amount.

yancyribbens: > That's the recipient output, not the change. The first output pays the recipient, so it's amount:

Ah got it. Thanks, that makes sense.

crates/wallet/src/silentpayments/sending.rs
```diff
@@ -0,0 +1,763 @@
...
+    Ok((selected, drain))
+}
+
+fn cs_feerate(fee_rate: FeeRate) -> bdk_coin_select::FeeRate {
+    bdk_coin_select::FeeRate::from_sat_per_wu(fee_rate.to_sat_per_kwu() as f32 / 1000.0)
+}
+
+fn output_weight(script_len: usize) -> Weight {
+    // 8-byte value, 1-byte length prefix, then the script, times 4 weight units per byte.
+    Weight::from_wu_usize((8 + 1 + script_len) * 4)
```
yancyribbens, line 360: I would probably not depend on `from_wu_usize` mostly because it is not available in `units` https://github.com/rust-bitcoin/rust-bitcoin/blob/master/units/src/weight.rs.  Hopefully units 1.0 is around the corner.

Anyway, assuming this is only for 32 and 64 bit architectures, id probably do something like

```rust
    let vb = u64::try_from(8 + 1 + script_len).unwrap();
    Weight::from_vb_unchecked(vb)
```

There should probably be some policy that says we don't guarantee no panic paths if someone is using an arch with more than 64 bits.

rustaceanrob:
Were you able to test that this works? I tried broadcasting a transaction and the CLI returns, but looking the logs (in `DEBUG`), I'm not seeing any `reject` messages, yet mempool.space is not picking it up and I haven't been included in a block.

It would be nice to have some transaction validation prior to broadcast so we know the broadcast code is buggy. Something like `testmempoolaccept`.

rustaceanrob:
Also probably good for people to know, you can get signet sats [here](https://silentpayments.dev/faucet/signet/). (please send them back once we figure this out)

pzafonte:
> Were you able to test that this works?

The testing I did prior was pushing a transaction through `broadcast-raw-tx`  on a live signet node, and the log confirmed it connected to a peer and sent the tx. I am testing propagation now with a valid transaction and checking mempool.space.

> It would be nice to have some transaction validation prior to broadcast so we know the broadcast code is buggy. Something like `testmempoolaccept`.

Good call, for right now I will add a check that verifies each signature against the output it spends before broadcast.

crates/wallet/src/silentpayments/sending.rs
yancyribbens: from_vb is what you actually want I believe (as commented bellow)

crates/wallet/src/silentpayments/sending.rs
yancyribbens: > I can use ChangePolicy::min_value_and_waste

here is what the README.md says about that `min_value_and_waste`.

> // We use a change policy that introduces a change output if doing so reduces
// the "waste" (i.e. adding change doesn't increase the fees we'd pay if we factor in the cost to spend the output later on).

That does sound like the correct behavior, since the ultimate goal of coin-selection is to reduce the waste metric.  Murch wrote a nice blog post about this: https://murch.one/posts/waste-metric/.

BTW, the waste metric to work "properly" needs to know the correct long_term_fee_rate.  Currently, it looks like you have that hard coded as: `const LONG_TERM_FEERATE_SAT_PER_VB: f32 = 1.0;`.  So, that will give you wildly inaccurate results in some cases because the algorithm will most likely always think it's in a high fee rate environment when in fact it may not be.  Hopefully that all makes sense.

crates/wallet/src/silentpayments/sending.rs
yancyribbens: And yes, that all tracks with me as well.  You'll have a changeless solution, which has only the `amount` sent to the recipient.  Then in the case of a change, there will be a second output that goes back to sender.  The address that goes back to sender looks like what's provided here ~`sp_output_script(&derived, sp)`~ `sp_output_script(&derived, change_address)?`   and the recipient goes to the scriptpub key here `address.script_pubkey()`?

crates/wallet/src/silentpayments/sending.rs
pzafonte: That does makes sense. Setting aside an implementation / integration of a fee estimator for a follow up, perhaps it then makes sense to do `long_term_fee_rate = current_fee_rate` so the waste metric is neutral rather than an assumed future.

yancyribbens: Yeah if it's not handy to find `long_term_fee_rate` now, some note to followup would be great.  In the very least maybe in the commit message.

yancyribbens: Any yes, `long_term_fee_rate = current_fee_rate` would at least be better than hardcoded 1.

crates/wallet/src/silentpayments/sending.rs
pzafonte: Yes. The recipient is `address.script_pubkey()` for a plain address, or `sp_output_script(&derived, sp)` for a silent payment one.

pzafonte:
What I'm finding is that `broadcast_transaction` opens a connection to a peer, sends the transaction, but closes the connection immediately. `send_message` writes the message to the socket, so closing right away tares down the connection before the peer is able to read, validate, and relay the transaction, and so it gets dropped before reaching the mempool.

Faucet payment into the wallet, received and scanned, block 309134:
https://mempool.space/signet/tx/9f506b3e4b5e32caa32665f422aecbd33f06707d1ca431079e8b7f152b71c45f

silently dropped on the immediate disconnect and then propagated once the connection was held open, block 309194:
https://mempool.space/signet/tx/79c6b5eb46516fabf467838fdb5d76918199b1884e81838540bd3ada635b836f

same but with a hack-y "fix" I was testing,  propagated and mined in block 309195:
https://mempool.space/signet/tx/b4d6acf43c833696a13c5e363375bd460c83b10580bf9115ed3d417edb8ccb7f

rustaceanrob:
> same but with a hack-y "fix" I was testing, propagated and mined in block 309195

What was the fix? I think we can do something like send a ping and wait for the pong. This is done in private broadcast to my knowledge.

pzafonte:
> > same but with a hack-y "fix" I was testing, propagated and mined in block 309195
> 
> What was the fix? I think we can do something like send a ping and wait for the pong. This is done in private broadcast to my knowledge.

I'm holding the connection open for a few seconds after sending so the peer can process and relay:
https://github.com/pzafonte/kernel-node/commit/da901d12798b57f1ab523c08a6123fb38269f353

Seems ping-pong might accomplish the same thing but in a more signal-driven way.

Two larger improvements that are not immediately pertinent, but are worth tracking separately 1) announcing the transaction before sending it or 2) relaying over the connection the node already keeps open. I can do a short write up in #51

rustaceanrob:
1/ is a nice-to-have but technically not required and I don't think would help in this problem 2/ was deliberately avoided. I plan on adding a Socks5 proxy so we can connect to peers over Tor, however we will likely want to do IBD without Tor, so the current design is flexible to broadcast transactions privately without slowing down IBD.

crates/wallet/src/silentpayments/sending.rs
yancyribbens: It would probably be best to just use a selection process that doesn't require knowledge of `long_term_fee_rate`.  If the goal is just to get something out there that works,  `SRD` (single random draw) would be simple, fast and doesn't require knowledge of cost_of_change nor long_term_fee_rate.  Then at some later point BnB and/or coin-grinder could be added.

pzafonte:
Makes sense on both. I started thinking along the lines of "how long do I keep the socket open." 

I implemented the ping-pong approach you described in #66 . 

Tested on signet, the spend propagated and confirmed in block 309302:
[https://mempool.space/signet/tx/06c111cd5f26ce92e6adff1a5329c053](https://mempool.space/signet/tx/06c111cd5f26ce92e6adff1a5329c0537c455029940b0701f7b633b43423ae0e)

rustaceanrob:
Can be rebased on #66

rustaceanrob:
Nice: https://mempool.space/signet/tx/9ec8598fd29b364f322eb78dc54ac79b47c01de622293b82f233362307ecb609

pzafonte:
For reference:
- As suggested, I changed coin selection to single random draw, which greatly simplifies things for the time being, and changed weight conversion to `from_vb_unchecked`.
- Each input's signature is now verified against the output it spends before the transaction is returned, which partially addresses pre-broadcast validation by catching faulty signing but not whether the transaction is otherwise valid in other ways.

pzafonte:
Yes, also tested it: https://mempool.space/signet/tx/002ea86e3a01d394f040c4dbf58ddd0f525cbe1257559d6aeb85cc1e3c2a6b34

Will send signet sats back. :)

rustaceanrob (approved):
I would like to continue moving things along. As a follow up we can:
- Implement RBF, this a big one because locked coins can be annoying
- Make the coin selection more robust.
- Add roundtrip tests with Bitcoin Core to make sure we're all square

Great work @pzafonte!

yancyribbens:
> As suggested, I changed coin selection to single random draw

It doesn't look like that's the case.  Also the LONG_TERM_FEE_RATE is still hard coded as `1.0`.

pzafonte:
Oh shoot, that push didn't go through. I pushed up on the same branch.

PR #68, opened 2026-06-18
silent payments: add sending support (correction)

pzafonte:
I had an intermittent network issue, and my last commits from #64 didn't get pushed up.

Restating this, but this commit should resolve the following from that PR:
- I changed coin selection to single random draw, which greatly simplifies things for the time being, and changed weight conversion to from_vb_unchecked.
- Each input's signature is now verified against the output it spends before the transaction is returned, which partially addresses pre-broadcast validation by catching faulty signing but not whether the transaction is otherwise valid in other ways.

### DIFF
--- a/crates/wallet/src/silentpayments/sending.rs
+++ b/crates/wallet/src/silentpayments/sending.rs
@@ -2,7 +2,6 @@ use std::collections::HashMap;
 use std::fmt;
 use std::str::FromStr;
 
-use bdk_coin_select::metrics::LowestFee;
 use bdk_coin_select::{
     Candidate, ChangePolicy, CoinSelector, Drain, DrainWeights, Target, TargetFee, TargetOutputs,
 };
@@ -22,8 +21,6 @@ use silentpayments::{Network, SilentPaymentAddress};
 
 use crate::silentpayments::wallet::{Coin, Wallet};
 
-const LONG_TERM_FEERATE_SAT_PER_VB: f32 = 1.0;
-
 #[derive(Debug)]
 pub enum SendError {
     WatchOnly,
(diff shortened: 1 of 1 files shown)


yancyribbens:
No problem, that's why we review.  Also, @rustaceanrob, if I happen to do a review, would it be possible to hold of on merging until I've had a chance to see if the changes are good?

I'll plan to look over these changes more tomorrow.  It's getting late here and I don't think this is super urgent tonight?

rustaceanrob:
Needs rebase

rustaceanrob:
>  if I happen to do a review, would it be possible to hold of on merging until I've had a chance to see if the changes are good?

Sure. The PR cycles should slow down now that we have the basis for a silent payments wallet. Also, I figured you may want to put your own pull requests up in your areas of interest. For now I think we should move on to testing.

yancyribbens:
Also, see: https://github.com/kernel-node/kernel-node/pull/70

yancyribbens:
> I changed coin selection to single random draw, which greatly simplifies things for the time being, and changed weight conversion to from_vb_unchecked.

FYI it looks like BDK is now adding SRD algo here: https://github.com/bitcoindevkit/coin-select/pull/50

yancyribbens:
Anyway, I think this can be closed in light of https://github.com/kernel-node/kernel-node/pull/77

pzafonte:
> Anyway, I think this can be closed in light of #77

It would supersede this in large part so I'll just stick this in draft status until #77 merges, and then close it out.

yancyribbens:
> Each input's signature is now verified against the output it spends before the transaction is returned, which partially addresses pre-broadcast validation by catching faulty signing but not whether the transaction is otherwise valid in other ways.

Not sure if you and Rob still want this change, but it could be a separate PR and commit.

PR #70, opened 2026-06-19
draft: use SRD coin-selection

yancyribbens:
I wanted to draft this instead of https://github.com/kernel-node/kernel-node/pull/68

Uses `SRD` which doesn't require knowledge of `long_term_fee_rate` or `cost_of_change` and is suitable for testing.  Can follow up with another PR to add `BnB` and `CoinGrinder` once we have more parameters.

If this looks ok, will do some cleanup and remove from draft state.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -62,10 +62,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "anyhow"
-version = "1.0.102"
+name = "arbitrary"
+version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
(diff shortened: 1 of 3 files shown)


yancyribbens:
rebased

yancyribbens:
let use https://github.com/kernel-node/kernel-node/pull/74 instead

yancyribbens:
Started working on some updates to coin-selection to make it more ergonomic.  Hopefully you guys don't mind, but I borrowed your naming convention of `Spendable` as the trait interface name: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/279.   The noun Spendable works much better than what I was using before (WeightedUtxo) I think.  Then consumers would impl like:

```rust
impl Spendable for Coin<'_> {
    fn total_weight(&self) -> Weight {}
    fn value(&self) -> Amount {}
}
```

PR #72, opened 2026-06-24
keys: add cli for address recovery

yancyribbens:
I forgot to write down the pub key after generation.  Thought there should be a way to go back and print it again later..

Also, it seems to me, the keys-file stuff should just be it's own crate under `/crates` that does the key generation and recovery.  That way, there isn't the need for the early `return`.

### DIFF
--- a/src/bin/cli.rs
+++ b/src/bin/cli.rs
@@ -69,6 +69,9 @@ enum WalletCmd {
         /// 32-byte x-only spend public key as hex.
         spend_key: String,
     },
+    PrintKeysFromKeysFile {
+        path: PathBuf,
+    },
     /// Show wallet balance.
     Balance,
     /// Show wallet transaction history.
@@ -141,6 +144,18 @@ fn main() {
         return;
     }
 
+    if let Commands::Wallet(WalletCmd::PrintKeysFromKeysFile { path }) = &cli.commands {
+        let read = SilentPaymentKeysFile::load(path)
+            .expect("file path provided should be readable as a silent payments keys file");
+
+        let spend_pub = read.spend_xonly();
+        eprintln!("spend_pub={}", spend_pub);
+
+        let scan_priv = read.scan_key;
+        eprintln!("scan_key={}", scan_priv.display_secret());
+        return;
+    }
+
     let rt = tokio::runtime::Builder::new_current_thread()
         .enable_all()
         .build()
@@ -174,6 +189,9 @@ fn main() {
                 let client = wallet_response.get().unwrap().get_wallet().unwrap();
                 match cmd {
                     WalletCmd::GenerateKeys { .. } => unreachable!("handled before runtime"),
+                    WalletCmd::PrintKeysFromKeysFile { .. } => {
+                        unreachable!("handled before runtime")
+                    }
                     WalletCmd::ImportKeys {
                         scan_key,
                         spend_key,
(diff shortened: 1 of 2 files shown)


src/bin/cli.rs
```diff
@@ -69,6 +69,9 @@ enum WalletCmd {
         /// 32-byte x-only spend public key as hex.
         spend_key: String,
     },
+    PrintAddressFromKeysFile {
```
rustaceanrob: This isn't an address in the conventional sense (there is a silent payments address that is different than this). I think you are looking for "PrintPublicKeysFromFile". I think there is no hurt in having this print both public keys as well.

yancyribbens: > I think there is no hurt in having this print both public keys as well.

They are the same.

for example

```rust
        let secp = Secp256k1::signing_only();
        let scan: SecretKey = read.scan_key;
        let scan_pub = PublicKey::from_secret_key(&secp, &scan);
        eprintln!("scan_pub={}", spend_pub);
```

prints the same Pubkey as spend_pub.

rustaceanrob: Sounds like a bug. These should be distinct keys.

yancyribbens: >  think you are looking for "PrintPublicKeysFromFile

Sounds good, updated.

yancyribbens: > Sounds like a bug. These should be distinct keys.

Could be.  I'd need to investigate this more.  I noticed that only the spend pub key is printed which is why in recovery I only print that one: https://github.com/kernel-node/kernel-node/blob/e56bb4d112f6c1cddbdf5c024e4084be20faab2c/src/bin/cli.rs#L138

rustaceanrob: True, the scan private key can be shared safely, just the spend private key needs to be kept secret. How about we print the scan _private key_ and spend _public key_. I think that would be useful.

yancyribbens: Sounds good.  updated to print scan priv key and spend pub key.

src/bin/cli.rs
```diff
@@ -141,6 +144,17 @@ fn main() {
         return;
     }
 
+    if let Commands::Wallet(WalletCmd::PrintKeysFromKeysFile { path }) = &cli.commands {
+        let read = SilentPaymentKeysFile::load(path).expect("valid keys file path");
```
rustaceanrob: Final nit here sorry. Can we change this exception to something more informative? i.e. `the path provided does not exist or is not a valid silent payments keys file`

rustaceanrob: Also instead of a panic which can produce word vomit maybe we use `if let Err` or a match to handle this case

rustaceanrob (approved):
This iteration does the job, thanks

src/bin/cli.rs
yancyribbens: > Final nit here sorry. Can we change this exception to something more informative? i.e. the path provided does not exist or is not a valid silent payments keys file

How about? `"file path provided should be readable as a silent payments keys file"`

> Also instead of a panic which can produce word vomit maybe we use if let Err or a match to handle this case

I think it should match what was done in `WalletCmd::GenerateKeys`.  So, if an Err is preferred, it should probably be changed both places.

yancyribbens:
> Sounds like a bug. These should be distinct keys.

There should probably be an issue created so this doesn't get lost..

PR #74, opened 2026-06-28
silent payments: use select coins algo

yancyribbens:
High level overview of this PR:
 
* Add discard_fee_rate to cli.
* Use select_coins function to make selection from rust-bitcoin-coin-selection lib

As a follow up, it would be nice to internally estimate both FeeRate and LongTermFeeRate using mempool and blockchain data similar to bitcoin core

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -68,6 +68,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
+name = "arbitrary"
+version = "1.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c3d036a3c4ab069c7b410a2ce876bd74808d2d0888a82667669f8e783a898bf1"
+
+[[package]]
 name = "arrayvec"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -75,19 +81,12 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "base58ck"
-version = "0.1.0"
+version = "0.1.101"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c8d66485a3a2ea485c1913c4572ce0256067a5377ac8c75c4960e1cda98605f"
+checksum = "365c0acd5b2e8dd0111a46c4faea83fb3cfb6e39a49a7c73a06e090db7b2eff0"
 dependencies = [
- "bitcoin-internals",
- "bitcoin_hashes 0.14.1",
+ "bitcoin_hashes 0.14.101",
 ]
-
-[[package]]
-name = "bdk_coin_select"
-version = "0.4.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43064fa3dd1d3a8b24079cfcb5ede6b785857edc782277f17a1736511ccc1916"
 
 [[package]]
 name = "bech32"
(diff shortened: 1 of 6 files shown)


yancyribbens:
marking as draft while I run some local tests.

yancyribbens:
Removed from draft state.

yancyribbens:
I just realized that another nice thing that bitcoin-coin-select has is the ability to define a max_transaction weight.  So BnB for example would not return a set that breaks the transaction weight max (400_000 wu).  bdk_coin_select does not seem to have the smarts to do that from what I can tell.

ref: https://docs.rs/bitcoin-coin-selection/0.8.5/bitcoin_coin_selection/fn.branch_and_bound.html

rustaceanrob:
I don't want this to be taken the wrong way in that I appreciate your work on coin selection, which is potentially an underdeveloped part of the ecosystem, but it appears you are the only active contributor to the crate being added. Can you remind me again why exactly some of these algorithms cannot be upstreamed to BDK? In the case of BDK, they have a foundation and handful of developers, where if one developer steps down, there is not much of a risk of a crate going stale. Given that they also have a broad base of users, not only is there a stronger maintenance guarantee, your work is more likely to be used in production wallets.

yancyribbens:
> I don't want this to be taken the wrong way in that I appreciate your work on coin selection, which is potentially an underdeveloped part of the ecosystem, but it appears you are the only active contributor to the crate being added

I agree, there is still a lot of work to develop this crate further.  Hopefully as bitcoin/units-1.0 becomes a thing, I will have advanced this crate to be less underdeveloped.  This also means finding more projects that I can help test with so as to know how best to build the API and tooling.

> Can you remind me again why exactly some of these algorithms cannot be upstreamed to BDK?

It would be possible with a lot of work for me to re-design the algorithm to work with BDK coin-select.  In viewing their project, it seems that the design patters are somewhat divergent from bitcoin-core from whom many of these selection algorithms originate.  I'm not saying it's not possible, it would just be a large time investment for me to re-write coin-grinder to fit.  Indeed I do think they should implement coin-grinder since it is the default algorithm that is used by bitcoin-core when fees are high.

As stated in the PR, since I've maintained this project since times before BDK, I think it is more advanced in many ways.  If you compare bench-marking times for example, this crate far outperforms BDK.  Not to mention, I have spent a lot of time to add fuzz tests, not to mention property tests.  All that is to say, I could port work into BDK, although, BDK could also just as easily start helping me out with this crate as well.

>  In the case of BDK, they have a foundation and handful of developers, where if one developer steps down, there is not much of a risk of a crate going stale.

I am the main contributor to this coin-selection crate, it is true.  For some time during the pandemic Tibo from crypto garage did all the maintenance, although, as far as I know, he has since moved to a different organization, and so is not directly involved anymore.  However, Murch helps with review from time to time, and he basically wrote the book on coin-selection as you may well know.  My hope is that as this gets further developed and gains adoption, I may have more help, however even large popular crates routinely end up being maintained by only a small number of people unfortunately.

> Given that they also have a broad base of users, not only is there a stronger maintenance guarantee, your work is more likely to be used in production wallets.

Bdk does have a broad user base even though as mentioned above, I feel like their crate has many gaps.  However, I do hope they have success in their project.  Often times its better for the ecosystem to have choices, and I'm happy to help with their project if I find a way that I think I could benefit.

yancyribbens:
After reflecting a bit, it sounds like there is not much enthusiasm for me adding this here based on the above conversation.  Thank you for allowing me the chance to do a bit of field testing with this crate though.  Will close for now.

yancyribbens:
Also I want to add I'm happy to help BDK, even if it's not of any benefit to me.  I think that came out wrong.  It's just that I have a lot of other things on my plate I want to work on, although if I can find a way to help their project I am happy to do so.

yancyribbens:
> In the case of BDK, they have a foundation and handful of developers, where if one developer steps down, there is not much of a risk of a crate going stale.

@rustaceanrob Just a thought, but if you and or @pzafonte wanted to help me maintain the coin-selection crate, we could probably make it a rust-bitcoin project.  It was originally intended to be moved to the rust-bitcoin org since it's inception.  However, there hasn't been anyone in rust-bitcoin that has the experience or time to help me out with this, so it never moved there.

PR #75, opened 2026-06-29
units: use u32 for fee rate

yancyribbens:
Use u32 to represent fee_rate since rust-bitcoin internally uses a `u64` to represent a fee-rate.  This also helps avoid a panic that can happen while multiplying by 250.  Also, may as well use rust-bitcoin to do the conversion from vB to kwu.

### DIFF
--- a/capnp/wallet.capnp
+++ b/capnp/wallet.capnp
@@ -6,5 +6,5 @@ interface Wallet {
     getHistory  @2 () -> (entries :Text);
     receive     @3 () -> (address :Text);
     broadcastRawTx @4 (tx :Data) -> (txid :Text);
-    sendToAddress @5 (address :Text, amountSat :UInt64, feeRateSatPerVb :Float64) -> (ok :Bool, message :Text);
+    sendToAddress @5 (address :Text, amountSat :UInt64, feeRateSatPerVb :UInt32) -> (ok :Bool, message :Text);
 }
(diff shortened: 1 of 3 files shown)


rustaceanrob:
Considering transactions can confirm recently for well under 1 sat/vb (0.1 sat/vb for instance), I don't know why we would do this

yancyribbens:
> Considering transactions can confirm recently for well under 1 sat/vb (0.1 sat/vb for instance), I don't know why we would do this

We are already doing it in the following:
```rust
 let fee_rate = FeeRate::from_sat_per_kwu((fee_rate_sat_per_vb * 250.0).ceil() as u64);
```

The problem with this is that multiply by 250 will overflow or wrap in release mode causing unpredictable results.

yancyribbens:
I think there is an argument that maybe the rust-bitcoin FeeRate type should not be `u64`?

rustaceanrob:
> The problem with this is that multiply by 250 will overflow or wrap in release mode causing unpredictable results.

The use of the float is such that we can express numbers below 1 sat/vb. If 1 sat/vb does not overflow (obviously it doesn't) neither will anything else than 1.

yancyribbens:
Hmm interesting.  I had thought `f64::MAX * 250.0` would overflow.  Appears I was wrong.

Thanks for the feedback.  Will close this then.

rustaceanrob:
I think it is fine to change this to a `Float32`, which should be more than enough to represent any reasonable fee rate, but we definitely don't want to change this to an integer

yancyribbens:
> I think it is fine to change this to a Float32, which should be more than enough to represent any reasonable fee rate, but we definitely don't want to change this to an integer

I'm not sure that there would be much advantage to that since f64::MAX * 250 seems be handled fine on 64bit machines.

PR #77, opened 2026-07-01
units: add consolidate_feerate as cli arg

yancyribbens:
The current behavior of hard-coding the long_term_fee_rate to be 1 can result in incorrect coin-selection behavior.

### DIFF
--- a/capnp/wallet.capnp
+++ b/capnp/wallet.capnp
@@ -6,5 +6,5 @@ interface Wallet {
     getHistory  @2 () -> (entries :Text);
     receive     @3 () -> (address :Text);
     broadcastRawTx @4 (tx :Data) -> (txid :Text);
-    sendToAddress @5 (address :Text, amountSat :UInt64, feeRateSatPerVb :Float64) -> (ok :Bool, message :Text);
+    sendToAddress @5 (address :Text, amountSat :UInt64, feeRateSatPerVb :Float64, consolidateFeeRateSatPerVb :Float64 = 10) -> (ok :Bool, message :Text);
 }
--- a/crates/wallet/src/silentpayments/sending.rs
+++ b/crates/wallet/src/silentpayments/sending.rs
@@ -22,8 +22,6 @@ use silentpayments::{Network, SilentPaymentAddress};
 
 use crate::silentpayments::wallet::{Coin, Wallet};
 
-const LONG_TERM_FEERATE_SAT_PER_VB: f32 = 1.0;
-
 #[derive(Debug)]
 pub enum SendError {
     WatchOnly,
@@ -143,6 +141,7 @@ impl Wallet {
         recipient: Recipient,
         amount: Amount,
         fee_rate: FeeRate,
+        long_term_fee_rate: FeeRate,
     ) -> Result<Transaction, SendError> {
         let spend_secret = self.spend_secret.ok_or(SendError::WatchOnly)?;
         let keys = self.keys.as_ref().ok_or(SendError::WatchOnly)?;
@@ -163,6 +162,7 @@ impl Wallet {
             recipient,
             amount,
             fee_rate,
+            long_term_fee_rate,
             self.scan_height,
             change_address,
             &coins,
@@ -170,11 +170,13 @@ impl Wallet {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 fn build_transaction(
     spend_secret: &SecretKey,
     recipient: Recipient,
     amount: Amount,
     fee_rate: FeeRate,
+    long_term_fee_rate: FeeRate,
     tip_height: u32,
     change_address: SilentPaymentAddress,
     coins: &[SpendableCoin],
@@ -213,10 +215,11 @@ fn build_transaction(
         DrainWeights::TR_KEYSPEND,
         change_dust.to_sat(),
         feerate,
-        bdk_coin_select::FeeRate::from_sat_per_vb(LONG_TERM_FEERATE_SAT_PER_VB),
+        cs_feerate(long_term_fee_rate),
     );
 
-    let (mut selected, drain) = select_coins(coins, target, change_policy, 100_000)?;
+    let (mut selected, drain) =
+        select_coins(coins, target, change_policy, long_term_fee_rate, 100_000)?;
     let mut rng = bitcoin::secp256k1::rand::thread_rng();
     selected.shuffle(&mut rng);
 
@@ -319,6 +322,7 @@ fn select_coins<'a, 'c>(
     coins: &'c [SpendableCoin<'a>],
     target: Target,
     change_policy: ChangePolicy,
+    long_term_fee_rate: FeeRate,
     max_bnb_rounds: usize,
 ) -> Result<(Vec<&'c SpendableCoin<'a>>, Drain), SendError> {
     let spendable: Vec<&SpendableCoin> = coins.iter().collect();
@@ -328,9 +332,10 @@ fn select_coins<'a, 'c>(
         .collect();
 
     let mut selector = CoinSelector::new(&candidates);
+    let long_term_feerate = cs_feerate(long_term_fee_rate);
     let metric = LowestFee {
         target,
-        long_term_feerate: bdk_coin_select::FeeRate::from_sat_per_vb(LONG_TERM_FEERATE_SAT_PER_VB),
+        long_term_feerate,
         change_policy,
     };
     if selector.run_bnb(metric, max_bnb_rounds).is_err() {
@@ -450,12 +455,14 @@ mod tests {
         }];
 
         let fee_rate = FeeRate::from_sat_per_vb(2).unwrap();
+        let long_term_fee_rate = FeeRate::from_sat_per_vb(10).unwrap();
         let amount = Amount::from_sat(50_000);
         let tx = build_transaction(
             &spend_secret,
             Recipient::SilentPayment(recipient),
             amount,
             fee_rate,
+            long_term_fee_rate,
             100,
             change_address,
             &coins,
@@ -510,6 +517,7 @@ mod tests {
             Recipient::SilentPayment(recipient),
             Amount::from_sat(20_000),
             FeeRate::from_sat_per_vb(2).unwrap(),
+            FeeRate::from_sat_per_vb(10).unwrap(),
             100,
             change_address,
             &coins,
@@ -540,12 +548,22 @@ mod tests {
         let amount = Amount::from_sat(50_000);
 
         let tx = wallet
-            .build_transaction(Recipient::SilentPayment(recipient), amount, fee_rate)
+            .build_transaction(
+                Recipient::SilentPayment(recipient),
+                amount,
+                fee_rate,
+                fee_rate,
+            )
             .unwrap();
         wallet.reserve_coins(tx.input.iter().map(|i| i.previous_output));
 
         let err = wallet
-            .build_transaction(Recipient::SilentPayment(recipient), amount, fee_rate)
+            .build_transaction(
+                Recipient::SilentPayment(recipient),
+                amount,
+                fee_rate,
+                fee_rate,
+            )
             .unwrap_err();
         assert!(matches!(err, SendError::NoSpendableCoins));
     }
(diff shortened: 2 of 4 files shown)


yancyribbens:
I added a commit to match more closely the conventions used in the core client (see commit message).

pzafonte:
For latent safety purposes would it make more sense to put the 10 sat/vB default at the ipc.rs boundary, with a named constant?

```
// mirrors bitcoin-core's DEFAULT_CONSOLIDATE_FEERATE (10 sat/vB)
const DEFAULT_CONSOLIDATE_FEE_RATE_SAT_PER_VB: f64 = 10.0;

let raw = p.get_consolidate_fee_rate_sat_per_vb();
if !raw.is_finite() || raw < 0.0 {
    return Err(capnp::Error::failed(
        "consolidate fee rate must be a non-negative number".to_string(),
    ));
}
let long_term_sat_per_vb = if raw == 0.0 {
    DEFAULT_CONSOLIDATE_FEE_RATE_SAT_PER_VB
} else {
    raw
};
```

yancyribbens:
> For latent safety purposes would it make more sense to put the 10 sat/vB default at the ipc.rs boundary, with a named constant?

Thanks for the suggestion, although, im not sure I fully understand.  Clap allows default arguments, so in the CLI options, there needs to be something done if the option is not provided I believe.  Are you suggesting changing the way the defaults are handled in cli.rs as well as the suggestion above?

I think the other part of your suggestion is that if a value is provided, and the value is 0.0, that it should be changed to a default value instead?  Not sure, I find that not so intuitive and maybe we should just error on a 0.0 consolidate_fee_rate.  However, it looks like currently 0.0 for fee-rate is accepted, which I think is actually a problem since the network will never propagate a transaction with a 0.0 feerate.

pzafonte:
Right now 0 is accepted.  My preference is to let the server own the default, which looks like how Core does it. Erroring on 0  makes sense but pulls the default back onto the CLI and capnp can't tell "sent 0" from "sent nothing," so once you reject 0 the client has to supply the value. That's what I was trying to avoid, so I lean server-side default.

yancyribbens:
> Right now 0 is accepted. My preference is to let the server own the default, which looks like how Core does it. Erroring on 0 makes sense but pulls the default back onto the CLI and capnp can't tell "sent 0" from "sent nothing," so once you reject 0 the client has to supply the value. That's what I was trying to avoid, so I lean server-side default.

It sounds like a good idea to enforce the default on the server.  However, I'm still a little unclear how the CLI should behave when it gets no argument, since it must then set a value which may as well be the default.  Never the less, I have added your suggestion as a fixup even though this now means the default is being set twice.  Please let me know if this is what you had in mind or what I am missing.

crates/wallet/src/silentpayments/sending.rs
```diff
@@ -213,10 +215,11 @@ fn build_transaction(
         DrainWeights::TR_KEYSPEND,
         change_dust.to_sat(),
         feerate,
-        bdk_coin_select::FeeRate::from_sat_per_vb(LONG_TERM_FEERATE_SAT_PER_VB),
+        bdk_coin_select::FeeRate::from_sat_per_vb(long_term_fee_rate.to_sat_per_vb_ceil() as f32),
```
pzafonte: `to_sat_per_vb_ceil()` rounds the fractional consolidate rates up (e.g. 2.5 to 3). Is that intentional, or should it use `cs_feerate(long_term_fee_rate)` ?

crates/wallet/src/silentpayments/sending.rs
```diff
@@ -330,7 +334,9 @@ fn select_coins<'a, 'c>(
     let mut selector = CoinSelector::new(&candidates);
     let metric = LowestFee {
         target,
-        long_term_feerate: bdk_coin_select::FeeRate::from_sat_per_vb(LONG_TERM_FEERATE_SAT_PER_VB),
+        long_term_feerate: bdk_coin_select::FeeRate::from_sat_per_vb(
+            long_term_fee_rate.to_sat_per_vb_ceil() as f32,
```
pzafonte: Same question as above.

crates/wallet/src/silentpayments/sending.rs
yancyribbens: No it was not intentional.  I didn't think super hard about his because it was meant as more of a bridge commit since a followup commit switched this all be native rust-bitcoin FeeRate.  That followup PR was closed so it's good you caught that.

I'm curious why bdk decided to create their own feerate types instead of just use rb types here.

crates/wallet/src/silentpayments/sending.rs
yancyribbens: reply above

yancyribbens:
If there's no further feedback I can autosquash  the commits then.  I don't think Github is smart enough to autosqush on merge.

src/bin/cli.rs
```diff
@@ -90,6 +88,11 @@ enum WalletCmd {
         amount_sat: u64,
         /// Fee rate, in satoshis per virtual byte.
         fee_rate_sat_per_vb: f64,
+        /// The maximum feerate in sats/vB at which transaction building may
+        /// use more inputs than strictly necessary so that the wallet's UTXO
+        /// pool can be reduced (default: 10 sats/vB).  Long term fee rate,
+        /// in satoshis per virtual byte.  Defaults to fee_rate if not supplied.
```
pzafonte: Says both 'default: 10' and 'Defaults to fee_rate if not supplied'.

src/bin/cli.rs
```diff
@@ -250,11 +238,20 @@ fn main() {
...
                         req.get().set_address(&address);
                         req.get().set_amount_sat(amount_sat);
                         req.get().set_fee_rate_sat_per_vb(fee_rate_sat_per_vb);
+
+                        if let Some(consolidate_fee_rate) = consolidate_fee_rate_sat_per_vb {
+                            req.get()
+                                .set_consolidate_fee_rate_sat_per_vb(consolidate_fee_rate);
+                        } else {
+                            req.get().set_consolidate_fee_rate_sat_per_vb(10.0);
+                        }
```
pzafonte: The server already maps 0.0 to default, so I think you can drop this else statement and it will fall through to the node's default.

src/ipc.rs
```diff
@@ -1,3 +1,5 @@
+const DEFAULT_CONSOLIDATE_FEE_RATE_SAT_PER_VB: f64 = 10.0;
```
pzafonte: nit: Put this below imports.

src/bin/cli.rs
yancyribbens: Unfortunately that doesn't seem to work.

If I change it to the following (dropping the else branch):
```rust
 req.get().set_consolidate_fee_rate_sat_per_vb(consolidate_fee_rate_sat_per_vb);
```

The compiler complains that it expected f64 but found Option<f64>.  If I change the cli to accept f64 instead of Option, then the parameter becomes a required param I believe.

src/ipc.rs
yancyribbens: lazy me, fixed.

src/bin/cli.rs
yancyribbens: oops, bad pasta copy.  fixed.

src/bin/cli.rs
pzafonte: Try just no else, when `None`, the setter is simply never called

```
if let Some(consolidate_fee_rate) = consolidate_fee_rate_sat_per_vb {
    req.get()
        .set_consolidate_fee_rate_sat_per_vb(consolidate_fee_rate);
}
```

yancyribbens: Oh interesting, that does work.  Huh, I forgot you could do if let some without an else I guess.

yancyribbens: marking resolved

crates/wallet/src/silentpayments/sending.rs
```diff
@@ -615,8 +648,9 @@ mod tests {
             .collect();
 
         let amount = Amount::from_sat(45_000);
+        let fee_rate = FeeRate::from_sat_per_vb(2).unwrap();
```
rustaceanrob: nit: seems like an unnecessary move

capnp/wallet.capnp
```diff
@@ -6,5 +6,5 @@ interface Wallet {
     getHistory  @2 () -> (entries :Text);
     receive     @3 () -> (address :Text);
     broadcastRawTx @4 (tx :Data) -> (txid :Text);
-    sendToAddress @5 (address :Text, amountSat :UInt64, feeRateSatPerVb :Float64) -> (ok :Bool, message :Text);
+    sendToAddress @5 (address :Text, amountSat :UInt64, feeRateSatPerVb :Float64, consolidateFeeRateSatPerVb :Float64) -> (ok :Bool, message :Text);
```
rustaceanrob: I think we use default syntax to define this feerate in capnp? See: https://capnproto.org/language.html#interfaces

I would have to read the rust docs to confirm

src/bin/cli.rs
```diff
@@ -90,6 +90,11 @@ enum WalletCmd {
         amount_sat: u64,
         /// Fee rate, in satoshis per virtual byte.
         fee_rate_sat_per_vb: f64,
+        /// The maximum feerate in sats/vB at which transaction building may
+        /// use more inputs than strictly necessary so that the wallet's UTXO
+        /// pool can be reduced (default: 10 sats/vB).  Long term fee rate,
+        /// in satoshis per virtual byte.
+        consolidate_fee_rate_sat_per_vb: Option<f64>,
```
rustaceanrob, line 104: I'm still lost as to what this value is for. Can we explain how this interacts with the `fee_rate_sat_per_vb` here? I think that is what most users would care about

yancyribbens: To add a little more to the docs comment, the consolidation fee (also known as long_term_fee_rate) helps to determine if the current fee_rate is high or low.  So, if fee_rate > consolidation_fee_rate, then the current fee_rate is high based on historical trends, and if consolidation_fee_rate > fee_rate then the current fee_rate level is considered cheap.  The reason this matters is that BnB will select more inputs to build a transaction when fees are cheap to _consolidate_ the utxo set.  Likewise, if the current environment is a low fee_rate environment, BnB will try to find a solution with the least number of UTXOs possible.

yancyribbens: That is why hard_coding the long_term_fee_rate to 1.0 is bad, because it will direct the selection algorithms to think that we are always in a high fee environment, since it's fairly likely that fee_rate is greater than 1 sat/vB.

yancyribbens: Also, by adding this as a parameter, its possible you could do a curl or wget request from some external API to find this value.

crates/wallet/src/silentpayments/sending.rs
yancyribbens: fair point.  Fixed.

capnp/wallet.capnp
yancyribbens: Do you mean feerate can be set using a default?  It doesn't look like defaults are set, although the docs show that there is a way to set a default value, for example:

`foo @0 :Int32 = 123;`

I guess we can define the default of `10` this way instead of the server side like @pzafonte suggested.  I'm fine eihter way, although it would be less code to have capnp set the default instead of it being set by the node.

rustaceanrob: Yeah that's what I mean, in the style of your example. I'm not sure how it works on the Rust side, but it would be very nice to do it that way because other apps can take advantage of that syntax. For example, Swift supports optional function arguments.

src/bin/cli.rs
rustaceanrob: Ok much clearer now. Would you be able to condense that a bit and add that logic in the doc comment?

capnp/wallet.capnp
yancyribbens: Add the default with cpnproto instead of server side.  Tested and it looks like it works.

src/bin/cli.rs
yancyribbens: Sure, done, with the caveat that this is how the bitcoin core selection works, as well as rust bitcoin coin-selection.  However, I'm reasonably sure that's how bdk coin-select works as well.  Although, as stated before, I'm not really as well versed on that project.

rustaceanrob (approved):
lgtm

PR #78, opened 2026-07-08
test: Add integration harness for the node binary

pzafonte:
Add a fixture that starts the node on regtest in a temporary datadir, drives it through the cli over the control socket, and stops it on drop.

Adds two tests that use the fixture. One checks that the node echoes over the control socket and stops on request. The other points the node at a regtest `bitcoind` started through `corepc-node` and waits for the daemon to report the incoming connection. 

`corepc-node` is pinned to the v30 client to match the Bitcoin Core daemon it talks to and uses its download feature so `bitcoind` is fetched automatically when it isn't already on `PATH` / `BITCOIND_EXE`

Followups:
#85 
#86

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -1,0 +1,102 @@
+#![allow(dead_code)]
+
+use std::path::{Path, PathBuf};
+use std::process::{Child, Command, Output};
+use std::time::{Duration, Instant};
+
+use corepc_node::{Conf, Node, P2P};
+
+const READY_TIMEOUT: Duration = Duration::from_secs(30);
+const STOP_TIMEOUT: Duration = Duration::from_secs(45);
+const POLL_INTERVAL: Duration = Duration::from_millis(50);
+
+const CLOSED_PEER: &str = "127.0.0.1:1";
+
+pub fn start_bitcoind() -> Node {
+    let exe = corepc_node::exe_path()
+        .expect("resolve bitcoind: downloaded build, or BITCOIND_EXE, or bitcoind on PATH");
+    let mut conf = Conf::default();
+    conf.p2p = P2P::Yes;
+    Node::with_conf(exe, &conf).unwrap()
+}
+
+pub struct TestNode {
+    process: Child,
+    datadir: PathBuf,
+    _tempdir: tempfile::TempDir,
+}
+
+impl TestNode {
+    pub fn start() -> Self {
+        Self::start_connected(CLOSED_PEER)
+    }
+
+    pub fn start_connected(peer: impl std::fmt::Display) -> Self {
+        let tempdir = tempfile::tempdir().unwrap();
+        let datadir = tempdir.path().canonicalize().unwrap();
+        let process = Command::new(env!("CARGO_BIN_EXE_node"))
+            .arg("--network")
+            .arg("regtest")
+            .arg("--datadir")
+            .arg(&datadir)
+            .arg("--connect")
+            .arg(peer.to_string())
+            .spawn()
+            .unwrap();
+        let node = Self {
+            process,
+            datadir,
+            _tempdir: tempdir,
+        };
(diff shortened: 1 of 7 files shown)


rustaceanrob:
Why are these in draft? Direction looks fine to me.

pzafonte:
> Why are these in draft? Direction looks fine to me.

Thanks! I still have it in draft because the tests don't follow the chain yet. I'm working on the next commit which will add  `getTip` plus a sync test that mines on Core and polls until they match... which I'm almost done with.

src/bin/cli.rs
```diff
@@ -34,6 +34,8 @@ enum Commands {
     Echo(Echo),
     /// Terminate the server.
     Stop,
+    /// Show the current chain tip.
+    Tip,
```
rustaceanrob: I would prefer if this was not a standalone command. Can we move this into a `Chain` variant with `ChainCmd` enum that holds chain related requests? You can also open that as a separate pull to clean this PR up.

tests/common/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+        let node = Self {
+            process,
+            datadir,
+            _tempdir: tempdir,
+        };
+        node.wait_until_ready();
+        node
+    }
+
+    pub fn cli(&self, args: &[&str]) -> Output {
```
rustaceanrob, line 55: While I do think we should be testing the CLI as well, I think for the core integration we can use capnp directly in the integration tests. A separate test can parse CLI output but I think it is prone to errors when trying to confirm we are in sync with Bitcoin Core.

src/bin/cli.rs
pzafonte: Done, Now in #85

tests/common/mod.rs
pzafonte: The sync/reorg tests now call `getTip` over the control socket.

tests/common/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    _tempdir: tempfile::TempDir,
+}
+
+impl TestNode {
+    pub fn start() -> Self {
+        Self::start_connected(CLOSED_PEER)
+    }
+
+    pub fn start_connected(peer: impl std::fmt::Display) -> Self {
+        let tempdir = tempfile::tempdir().expect("create temp datadir");
```
rustaceanrob: I think the usual guidance for when to use `expect` vs `unwrap` is when the `expect` output would definitely provide better context than `unwrap`. For third party crates, we should opt for `unwrap` and use their documentation to debug. I would prefer if the rest of the `expect` were converted to `unwrap` as well in this PR, as they may contain OS specifc information (like spawning a process for instance).

tests/common/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+
+const READY_TIMEOUT: Duration = Duration::from_secs(30);
+const STOP_TIMEOUT: Duration = Duration::from_secs(45);
+const POLL_INTERVAL: Duration = Duration::from_millis(50);
+
+const CLOSED_PEER: &str = "127.0.0.1:1";
+
+pub fn start_bitcoind() -> Node {
+    let exe = corepc_node::exe_path()
+        .expect("resolve bitcoind: downloaded build, or BITCOIND_EXE, or bitcoind on PATH");
```
rustaceanrob, line 17: This `expect` looks good to me though.

tests/common/mod.rs
rustaceanrob: I went ahead and updated these for you

rustaceanrob:
I want to help get these changes in so I force pushed with a few changes:
- made the CI to also run the integration tests
- reduced the loop time on the feeler connections so the thread exits faster. The feeler interval is still the same.
- migrated `expect` to `unwrap` in most cases

PR #79, opened 2026-07-08
Add wallet CLI

yancyribbens:
Remove commands from node cli which would otherwise require an early return and are `unreachable`.  Logically these commands belong to wallet, so create wallet cli, and move commands to be under wallet instead of node.

Also, I think this would make the wallet more extensible in that we could start adding more key types and functionality.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1140,6 +1140,7 @@ dependencies = [
  "bdk_coin_select",
  "bitcoin",
  "bitcoinkernel",
+ "clap",
  "log",
  "silentpayments",
  "tempfile",
--- a/crates/wallet/Cargo.toml
+++ b/crates/wallet/Cargo.toml
@@ -9,6 +9,10 @@ bitcoin = { version = "0.32.8", features = ["rand-std"] }
 bitcoinkernel = "0.2"
 log = "0.4"
 bdk_coin_select = "0.4.1"
+clap = { version = "4", features = ["derive"]  }
 
 [dev-dependencies]
 tempfile = "3"
+
+[[bin]]
+name = "cli"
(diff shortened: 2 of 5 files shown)


rustaceanrob:
I would personally find it confusing to have different CLI binaries for different actions. These are already subspaced by `cli wallet ...`. I don't really follow why this is needed

yancyribbens:
> I would personally find it confusing to have different CLI binaries for different actions. These are already subspaced by cli wallet .... I don't really follow why this is needed

I'm fine to close this if you don't find it an interesting direction.  Personally though, I think the wallet should be decoupled from the node implementation such that you could run any wallet binary on a machine separate from the node.  This is sort of a step in that direction.

The other nice thing about decoupling the node and the wallet is that would allow each to be packaged and installed on crates.io separately and installed separately.  Assuming these binaries are to be available on crates.io of course.

rustaceanrob:
Considering the CLI can already be installed without the node binary, and the commands broken out here don't require a node to be running, I don't think there is much motivation to do this. I do think we should not require a node to be running for many wallet operations (signing, key generation), but I don't think that should be a separate CLI.

yancyribbens:
Ok, so what you are saying is that an end-user looking to install this would a) install node b) install cli c) install wallet all as separate binaries?  That seems fine, although wouldn't the CLI belong then in the crates folder since it's sort of a separate package it sounds like?

rustaceanrob:
No, I am saying installing the CLI already gives you the commands here without having to run the node at all. Any user can install the CLI on whatever machine they want and generate keys without installing the node binary.

yancyribbens:
Alright, I feel like I may not completely understand the end goal here so this PR might be not the correct direction.  I'll close this and cherry-pick other small changes.

PR #81, opened 2026-07-14
refactor: use match statement for wallet cli commands

yancyribbens:
Instead of multiple "if let" statements, this can instead be combined into a single match statement


yancyribbens:
Apologies, it doesn't look like this refactor makes sense to do at this point, closing for now.

PR #82, opened 2026-07-15
change key visibility modifier

yancyribbens:
Change key visibility modifier to be private.  Also, use similar naming convention among keys.


pzafonte:
I think you intended to dropping `pub` from the two fields, so it should be like this? 
```
pub struct SilentPaymentKeysFile {
    scan_key: SecretKey,
    spend_key: SpendKey,
}
```

yancyribbens:
> I think you intended to dropping pub from the two fields, so it should be like this?

Oh my,  I cherry-picked this from a different branch and failed.  Thanks for catching that.

rustaceanrob:
Looks fine from a code point of view. Can you elaborate on the motivation?

yancyribbens:
its better not to expose the internal field representation and instead use an access modifier since this needs to be exposed externally to users of the api.

rustaceanrob (approved):
I think a great thing about Rust is the data oriented nature of the language, but in this case I think it makes sense as the keys file may change in the future

PR #85, opened 2026-07-19
node: Add a getTip RPC for the chain tip

pzafonte:
Nothing on the control socket reports the node's chain position, so tests can see it peer with Bitcoin Core but not that it followed the chain. Add `getTip`, returning the active chain height and best block hash, and a chain tip subcommand to the cli to print them.

Depends on #78 
Followup #86

### DIFF
--- a/capnp/server.capnp
+++ b/capnp/server.capnp
@@ -1,9 +1,11 @@
 @0xc2e20cc9503cf68f;
 
 using Wallet = import "wallet.capnp";
+using Chain = import "chain.capnp";
 
 interface Server {
     echo @0 (msg :Text) -> (reply :Text);
     shutdown @1 () -> ();
     makeWallet @2 () -> (wallet :Wallet.Wallet);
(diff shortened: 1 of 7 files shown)


rustaceanrob:
Merged #78

capnp/server.capnp
```diff
@@ -6,4 +6,5 @@ interface Server {
     echo @0 (msg :Text) -> (reply :Text);
     shutdown @1 () -> ();
     makeWallet @2 () -> (wallet :Wallet.Wallet);
+    getTip @3 () -> (height :UInt32, hash :Text);
```
rustaceanrob: I should've made this comment along with the CLI, but similarly I think we will want a `chain.capnp` schema similar to the wallet. I think we will inevitably add things like reading blocks from disk and so on, so it seems prudent to add now.

rustaceanrob (approved):
LGTM

PR #86, opened 2026-07-19
test: Verify the node tracks Bitcoin Core's chain

pzafonte:
Add two tests that drive the node against a regtest `bitcoind` through `getTip`. One mines blocks and waits for the node to reach the same tip. The other invalidates a block on Core and mines a longer competing branch, then waits for the node to reorg to it. The competing branch is mined to a fresh address so its blocks differ from the invalidated ones.

Depends on #85

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -5,10 +5,13 @@ use std::process::{Child, Command, Output};
 use std::time::{Duration, Instant};
 
 use corepc_node::{Conf, Node, P2P};
+use kernel_node::server_capnp::server;
+use tokio_util::compat::{TokioAsyncReadCompatExt, TokioAsyncWriteCompatExt};
 
 const READY_TIMEOUT: Duration = Duration::from_secs(30);
 const STOP_TIMEOUT: Duration = Duration::from_secs(45);
 const POLL_INTERVAL: Duration = Duration::from_millis(50);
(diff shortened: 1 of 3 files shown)


rustaceanrob:
Needs rebase

pzafonte:
> Needs rebase

Rebased and changed some `.expect` to `.unwrap()`

Temporarily stacking a commit to test the macOS runner. Will remove it later.

pzafonte:
The intermittent macOS CI failure might be a shutdown race in the node.

tests/common/mod.rs
```diff
@@ -61,6 +81,53 @@ impl TestNode {
             .unwrap()
     }
 
+    pub fn tip(&self) -> (u32, bitcoin::BlockHash) {
+        let socket = self.socket_path();
+        let rt = tokio::runtime::Builder::new_current_thread()
```
rustaceanrob: Should we hold this runtime and initialize it once at `TestNode` startup? Seems like a non-negligible cost that might be avoidable.

rustaceanrob:
> The intermittent macOS CI failure might be a shutdown race in the node.

Please open an issue whenever you are available with the debug details. Otherwise things are looking good here.

tests/common/mod.rs
pzafonte: Good call, looks like it.

rustaceanrob (commented):
LGTM. Tested a handful of times locally

PR #87, opened 2026-07-22
test: verify the node receives a silent payment

pzafonte:
Core cannot send to a silent payment address, so the test builds the BIP-352 payment itself from a P2WPKH coinbase output, broadcasts it through Core, and mines it. The node, connected to Core, scans the block and its balance reflects the payment.

Adds helpers to import keys and read the receive address and balance over the control socket, and silentpayments as a dev-dependency to build the payment.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -4,8 +4,18 @@ use std::path::{Path, PathBuf};
 use std::process::{Child, Command, Output};
 use std::time::{Duration, Instant};
 
+use bitcoin::hashes::Hash;
+use bitcoin::secp256k1::{Message, Secp256k1, SecretKey};
+use bitcoin::sighash::{EcdsaSighashType, SighashCache};
+use bitcoin::{
+    absolute::LockTime, key::TweakedPublicKey, transaction::Version, Address, Amount,
+    CompressedPublicKey, Network, OutPoint, ScriptBuf, Sequence, Transaction, TxIn, TxOut, Witness,
+};
 use corepc_node::{Conf, Node, P2P};
 use kernel_node::server_capnp::server;
+use silentpayments::sending::generate_recipient_pubkeys;
+use silentpayments::utils::sending::calculate_partial_secret;
+use silentpayments::SilentPaymentAddress;
 use tokio_util::compat::{TokioAsyncReadCompatExt, TokioAsyncWriteCompatExt};
 
 const READY_TIMEOUT: Duration = Duration::from_secs(30);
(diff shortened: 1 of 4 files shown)


rustaceanrob:
Needs a rebase

tests/common/mod.rs
```diff
@@ -131,6 +141,76 @@ impl TestNode {
...
+                    .get()
+                    .unwrap()
+                    .get_address()
+                    .unwrap()
+                    .to_string()
+                    .unwrap()
+            }))
+    }
+
+    pub fn balance(&self) -> u64 {
```
rustaceanrob: Perhaps we return this as `Amount`

tests/common/mod.rs
```diff
@@ -170,3 +250,91 @@ impl Drop for TestNode {
...
+    let secp = Secp256k1::new();
+
+    let sender_sk = SecretKey::from_slice(&[0x21; 32]).unwrap();
+    let sender_pk = CompressedPublicKey(sender_sk.public_key(&secp));
+    let sender_address = Address::p2wpkh(&sender_pk, Network::Regtest);
+
+    // 101 blocks so the first coinbase is mature, in small batches so a
+    // connected node stays near the tip instead of rejecting a burst.
+    for _ in 0..101 {
+        core.client.generate_to_address(1, &sender_address).unwrap();
```
rustaceanrob: Can't we just use `generate_to_address` with `101`?

yancyribbens:
> Core cannot send to a silent payment address,

Is this just a temporary solution until core can send to a silent payment address?  I see there are some PRs to add SilentPayments to core although I haven't inspected them closely.

rustaceanrob:
> Is this just a temporary solution until core can send to a silent payment address? I see there are some PRs to add SilentPayments to core although I haven't inspected them closely.

Silent payments sending definitely won't make it into the October release, so we're looking at 8-9 months minimum. The sending code is pretty readable here as well so I think it's fine.

rustaceanrob:
Tested 5 times and passed each time with no intermittent failures. Nice job.

tests/wallet_receive.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use bitcoin::hex::DisplayHex;
+use bitcoin::secp256k1::{Secp256k1, SecretKey};
+use bitcoin::Amount;
+use common::{fund_silent_payment, start_bitcoind, TestNode};
+
+const SYNC_TIMEOUT: Duration = Duration::from_secs(60);
+
+#[test]
+fn receives_a_silent_payment() {
```
yancyribbens, line 13: I think it's a bad idea to add this as a unit test.  This is by definition an integration test and should exist separately from unit tests.

yancyribbens:
Running`cargo test` I get the following error out of the box:

```
test connects_to_bitcoin_core ... FAILED
 
failures:
 
---- connects_to_bitcoin_core stdout ----
 
thread 'connects_to_bitcoin_core' (61607) panicked at tests/common/mod.rs:228:9:
node did not exit within 45s after stop
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
 
 
failures:
    connects_to_bitcoin_core

```

IIUC I don't think running a bitcoin core node should be required to run the unit test suit.

rustaceanrob:
It isn't a unit test, integration tests run by default. `cargo test --lib --workspace` is what you're looking for.

yancyribbens:
That also fails for me.

> It isn't a unit test, integration tests run by default. cargo test --lib --workspace is what you're looking for.

```
test follows_bitcoin_core_chain ... FAILED
 
failures:
 
---- follows_bitcoin_core_chain stdout ----
 
thread 'follows_bitcoin_core_chain' (62153) panicked at tests/common/mod.rs:228:9:
node did not exit within 45s after stop
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
 
 
failures:
    follows_bitcoin_core_chain
```

rustaceanrob:
`BITCOIND_EXE` needs to be set with a path to Bitcoin Core, if it isn't already. FWIW that also seems to be an intermittent failure in this project we need to sort out.

I think it is maybe an annoyance at worst for us to have some extra steps for testing, but given Bitcoin Core is the dominant implementation and this is a node software, I think it is prudent we have tests for these scenarios.

tests/common/mod.rs
pzafonte: Done. `balance()` and `wait_for_balance()` now return `Amount`.

tests/common/mod.rs
pzafonte: Done. I was worried about the peer stalling, but it seems to be fine.

pzafonte:
> Silent payments sending definitely won't make it into the October release, so we're looking at 8-9 months minimum.

corepc-node caps us at Core 30.2 anyway, so this would also need a silent-payments-capable Core to be available through corepc-node as well

rustaceanrob:
tACK b25fdf1f5fd7f96290c2a3314ebbf646f213eba7

@yancyribbens if you have a different test layout suggestion please open an issue and I'll review

yancyribbens:
Interesting, I didn't realize "cargo test" would run both unit and integration tests.  I wonder if it would be worth documenting how to run wallet unit tests and also run unit tests separately in CI.  I guess for now I can just "cd ./crates/wallet && cargo test" to run only the wallet unit tests.

rustaceanrob:
https://doc.rust-lang.org/cargo/commands/cargo-test.html#description:

> Compile and execute unit, integration, and documentation tests.

The `cargo test` command runs the entire suite, including integration tests and doc tests.

`cargo test --lib` is the unit test flag. I think a markdown on testing in the `doc` folder would be useful, since we have the added requirement of a bitcoin core binary.

yancyribbens:
> The cargo test command runs the entire suite, including integration tests and doc tests.

Yeah I see that now, thanks.

> cargo test --lib is the unit test flag. I think a markdown on testing in the doc folder would be useful, since we have the added requirement of a bitcoin core binary.

This doesn't run any tests as you may have noticed.  It doesn't look like the node has any unit tests which I think is the reason.

rustaceanrob:
Yeah, not sure why but needs the extra `workspace` flag. We can start to document this stuff for sure.

yancyribbens:
> Yeah, not sure why but needs the extra workspace flag. We can start to document this stuff for sure.

Oh, nice that works also with workspace: `cargo test --lib --workspace`.  I though I had tried that before and it failed but I must have been mistaken.

PR #88, opened 2026-07-24
test: verify the node's transaction is valid to Core

pzafonte:
Fund a spend-capable node, have it build and broadcast a send to a Core
address, and confirm Core accepts the node-built transaction into its
mempool. Gives `start_connected` an optional keys argument and a
`random_signing_keys` helper so the test can start a signing-capable
node, and adds a helper that polls Core's mempool for the `txid`.

Adds a separate commit to verify the wallet imports keys over the CLI.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -17,6 +17,8 @@ use silentpayments::sending::generate_recipient_pubkeys;
 use silentpayments::utils::sending::calculate_partial_secret;
 use silentpayments::SilentPaymentAddress;
 use tokio_util::compat::{TokioAsyncReadCompatExt, TokioAsyncWriteCompatExt};
+use wallet::io::FileExt;
+use wallet::silentpayments::{SilentPaymentKeysFile, SpendKey};
 
 const READY_TIMEOUT: Duration = Duration::from_secs(30);
 const STOP_TIMEOUT: Duration = Duration::from_secs(45);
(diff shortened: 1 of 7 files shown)


tests/common/mod.rs
```diff
@@ -63,17 +65,35 @@ impl TestNode {
...
             .arg("--network")
             .arg("regtest")
             .arg("--datadir")
             .arg(&datadir)
             .arg("--connect")
-            .arg(peer.to_string())
-            .spawn()
-            .unwrap();
+            .arg(peer.to_string());
+        if spender {
```
rustaceanrob: Given there is very little cost to doing this unconditionally, I think we can omit the `spender` condition and simply have every test node instance have a wallet available. Contributors can write spending or receiving tests without configuration.

The previous test used the CLI to import the keys. Perhaps we have a separate test that validates the CLI import path, and use the startup option by default?

You may also ignore this feedback. I'm just hesitant whenever I see pure `if` condition. Another option is to have `start_connected` take an optional set of keys or similar. That way test writers don't have to decide between two startup options

pzafonte: That's fair and makes sense. I'll switch to an optional set of keys on `start_connected` and pass them in, plus a dedicated test for the CLI import path.

rustaceanrob:
Looks good to me. Why is this in draft?

pzafonte:
> Why is this in draft?

I had it that way because the CI was red before. I'm debugging that issue today.

rustaceanrob (approved):
lgtm thanks for closing the loop here

PR #90, opened 2026-07-29
node: Exit the peer thread promptly on shutdown

### LINKED ISSUE #89: Node hangs on shutdown when a peer connects after stop is signalled
If the peer thread finishes connecting after the shutdown flag is set, it misses the flag and blocks in the read loop until the 60s p2p read timeout instead of exiting. Shows up in CI as intermittent `node did not exit within 45s after stop`.

Set up a mock peer that holds the handshake open until stop, then completes it -> [here](https://github.com/pzafonte/kernel-node/tree/debug-shutdown)
```
INFO  net   Connected to Address { ... }
INFO  node  DBG net: connected, stored writer, running=false
panicked: node did not exit within 20s
```
`running` is already `false` when the thread finishes connecting, but nothing between the handshake and the read loop re-checks it.

I believe the expected behavior should be every worker observes the flag and exits within ~1–2s regardless of connection timing.

pzafonte:
A reconnect racing shutdown blocks on a new connection until the 60s p2p timeout. Set running to false before closing the peer connection, and re-check it after connecting. Also give the addr thread a receive timeout so it stops on the flag.

Closes #89

### DIFF
--- a/src/bin/node.rs
+++ b/src/bin/node.rs
@@ -366,6 +366,9 @@ fn run(
                     continue;
                 }
             };
+            if !running_peer.load(Ordering::SeqCst) {
+                break;
+            }
             loop {
                 if let Err(e) = peer.receive_and_process_message(&mut node_state) {
                     match e {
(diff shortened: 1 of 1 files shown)


rustaceanrob:
Looks good, thought all threads followed this pattern but good catch.

PR #92, opened 2026-07-31
Add setup wizard binary

### LINKED ISSUE #91: Add startup wizard
I did a demo of this repo and realized I forgot to use the `--sp-keys-file` startup option. We have a handful of other options like data directories and so on. I think a simple binary that walks through the options and prompts the user to select their preferences

Something like:
```
> cargo run --bin wizard

which network would you like to run on?
> signet
   bitcoin 

where should data be stored?
> ~/.kernel-node

would you like to import keys?
```

rustaceanrob:
Closes #91 

Common configuration options can be set via this setup program that prompts the user. This uses the existing `libc` dependency for raw mode in the terminal.

### DIFF
--- a/src/bin/wizard.rs
+++ b/src/bin/wizard.rs
@@ -1,0 +1,260 @@
+use std::borrow::Cow;
+use std::io::{self, BufRead, Read, Write};
+use std::os::fd::AsRawFd;
+use std::process::Command;
+
+use kernel_node::ext::ExpandTildeExt;
+
+const RESET: &str = "\x1b[0m";
+const BOLD: &str = "\x1b[1m";
+const DIM: &str = "\x1b[2m";
+const CYAN: &str = "\x1b[36m";
+const GREEN: &str = "\x1b[32m";
+const YELLOW: &str = "\x1b[33m";
+
+fn read_line() -> String {
+    let mut line = String::new();
+    io::stdin()
+        .lock()
+        .read_line(&mut line)
+        .expect("failed to read stdin");
+    line.trim().to_string()
+}
+
+fn prompt_with_default(question: &str, default: &str) -> String {
+    print!("{question} {DIM}[{default}]{RESET}: ");
+    io::stdout().flush().unwrap();
+    let answer = read_line();
+    if answer.is_empty() {
+        default.to_string()
+    } else {
+        answer
+    }
+}
+
+fn prompt_optional(question: &str) -> Option<String> {
+    print!("{question} {DIM}(enter to skip){RESET}: ");
+    io::stdout().flush().unwrap();
+    let answer = read_line();
+    if answer.is_empty() {
+        None
+    } else {
+        Some(answer)
+    }
+}
+
+fn prompt_yes_no(question: &str, default: bool) -> bool {
+    let hint = if default { "Y/n" } else { "y/N" };
+    loop {
+        print!("{question} {DIM}[{hint}]{RESET}: ");
+        io::stdout().flush().unwrap();
+        let answer = read_line();
+        if answer.is_empty() {
+            return default;
+        }
+        match answer.to_ascii_lowercase().as_str() {
+            "y" | "yes" => return true,
+            "n" | "no" => return false,
+            _ => println!("please answer y or n"),
+        }
+    }
+}
+
+struct RawMode {
+    fd: i32,
+    original: libc::termios,
+}
+
+impl RawMode {
+    fn enter(fd: i32) -> Option<Self> {
+        unsafe {
+            let mut original: libc::termios = std::mem::zeroed();
+            if libc::tcgetattr(fd, &mut original) != 0 {
+                return None;
+            }
+            let mut raw = original;
+            raw.c_lflag &= !(libc::ICANON | libc::ECHO | libc::ISIG);
+            raw.c_cc[libc::VMIN] = 1;
+            raw.c_cc[libc::VTIME] = 1;
+            if libc::tcsetattr(fd, libc::TCSANOW, &raw) != 0 {
+                return None;
+            }
+            Some(RawMode { fd, original })
+        }
+    }
+}
+
+impl Drop for RawMode {
+    fn drop(&mut self) {
+        unsafe {
+            libc::tcsetattr(self.fd, libc::TCSANOW, &self.original);
+        }
+        print!("\x1b[?25h");
+        io::stdout().flush().ok();
+    }
+}
+
+/// Arrow-key selector. Falls back to numeric input if stdin isn't a tty.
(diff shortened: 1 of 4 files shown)


src/bin/wizard.rs
```diff
@@ -0,0 +1,233 @@
...
+        args.push("--sp-keys-file".into());
+        args.push(k);
+    }
+    if daemon {
+        args.push("--daemon=true".into());
+    }
+
+    println!();
+    println!("{YELLOW}Run:{RESET}");
+    println!("  {GREEN}cargo {}{RESET}", args.join(" "));
```
pzafonte: A datadir with spaces silently runs against a truncated path. Perhaps quote each arg with a single-quote?

src/bin/wizard.rs
```diff
@@ -0,0 +1,233 @@
...
+        "Which network would you like to run on?",
+        &["signet", "bitcoin"],
+        0,
+    );
+
+    let datadir = prompt_with_default("Where should data be stored?", "~/.kernel-node");
+
+    let connect = prompt_optional("Connect only to a specific node (ip:port)?");
+
+    let sp_keys_file = prompt_optional("Path to a silent payments keys file to import?");
```
pzafonte: Some relative key paths fail here.

`keys.bin` fails when paired with `--daemon=true`

` ~/keys.bin` fails with or without the flag.

pzafonte (commented):
I mostly thew some odd inputs at the prompts. 
 
 Otherwise, good to have a wizard guided setup!

src/bin/wizard.rs
rustaceanrob: I think the daemon and relative path combo is broken right now regardless because of the forking logic, not sure of a way around this.

rustaceanrob: Internally the wizard can attempt to expand all paths to full rather than relative. That should fix it before the process fork

rustaceanrob: Expanding paths before executing the binary 4ebde349bc91efe5315bcfccfd873acaecf896e5. `key.bin` in the working directory or paths like `~/keys.bin` are working for me now.

src/bin/wizard.rs
rustaceanrob: Passing with `~/My Node` is creating the `'My Node'` folder in my home directory. Maybe this is a Linux/MacOS discrepancy in the `PathBuf` type?

pzafonte: Apologies, cold card related stuff ate my day yesterday so I didn't get to suggesting fixes. Also my comment wasn't clear on this, I meant the printed string, not the `args` passed to `Command`.  Typed bare, `~/My Node` works when the wizard launches the node itself, since `Command::args` passes it as one `argv` entry, it's only the printed line that splits when pasted. It's also worth noting that path expansion can also introduce spaces the user never typed.

Here's a suggestion...

```suggestion
let printable: Vec<Cow<'_, str>> = args.iter().map(|a| shell_quote(a)).collect();
println!("  {GREEN}cargo {}{RESET}", printable.join(" "));
```

```
use std::borrow::Cow;

fn shell_quote(arg: &str) -> Cow<'_, str> {
    let safe = |c: char| c.is_ascii_alphanumeric() || "-_./=:,@+~".contains(c);
    if !arg.is_empty() && arg.chars().all(safe) {
        Cow::Borrowed(arg)
    } else {
        Cow::Owned(format!("'{}'", arg.replace('\'', r"'\''")))
    }
}
```

Each `arg` is printed bare if it's made only of characters a shell already treats literally, otherwise I am wrapping them in single quotes:
```
signet                     -> signet
--daemon=true              -> --daemon=true
~/.kernel-node             -> ~/.kernel-node
/tmp/kn demo               -> '/tmp/kn demo'
/tmp/$HOME/keys.bin        -> '/tmp/$HOME/keys.bin'
/home/me/o'brien/keys.bin  -> '/home/me/o'\''brien/keys.bin'
```
I put single quotes on something like $HOME so it stays literal instead of expanding.  I put `~` in the bare set because quoting it would suppress expansion and make the default datadir look broken, but that's a judgement call on my part, I think it's correct either way.  The last one is an awkward case. You can't nest a single quote, so it closes, emits an escaped one, and reopens.

src/bin/wizard.rs
pzafonte: Great! I was thinking about path expansion, but wasn't sure about it.

src/bin/wizard.rs
rustaceanrob: Nice, applied that suggestion in the latest push. Sorry about the coldcard debacle. In fact, I was gonna open and issue to potentially bump our `rand` and make explicit we are using the `SysRng`, which internally uses the `getrandom` syscall on the Linux kernel. This syscall uses entropy generated from different sources like CPU temperature, mouse clicks, etc and is believed to be CSRNG

pzafonte: > In fact, I was gonna open and issue to potentially bump our `rand` and make explicit we are using the `SysRng`, which internally uses the `getrandom` syscall on the Linux kernel. This syscall uses entropy generated from different sources like CPU temperature, mouse clicks, etc and is believed to be CSRNG

Agreed, was thinking about that & related matters as well.

pzafonte:
ACK 76a809d

src/bin/wizard.rs
yancyribbens: >  I was gonna open and issue to potentially bump our rand and make explicit we are using the SysRng

which crate are you referring to?  I noticed the wallet uses rand via secp:
`let mut rng = bitcoin::secp256k1::rand::thread_rng();`

I don't see rand in the node crate though..

rustaceanrob: Yeah, `rand` is the correct crate, just an older version. `thread_rng` is considered secure. The propose of the refactor would be to make it abundantly clear to ourselves and others that the source is CSRNG. The current state of the [rand book](https://rust-random.github.io/book/guide-rngs.html#cryptographically-secure-pseudo-random-number-generators-csprngs) [recommends](https://docs.rs/rand/latest/rand/rngs/struct.StdRng.html#seeding-construction):

```rust
let rng = StdRng::try_from_rng(&mut SysRng).unwrap();
```

yancyribbens: Hmm I created this PR last year (in october) to bump rand from 0.8 to 0.9 https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/234 for the coin-selection crate.  Cant remember why, but the PR switches from thread_rng to rng it looks like.

yancyribbens: Anyway, I think you would want to open an issue on the secp256k1 crate to bump the rand dep there?  Although it looks like master is on 0.9.x for rand.

PR #93, opened 2026-08-13
test: Verify a reorg drops a payment that is not re-mined

pzafonte:
Invalidate a block holding a silent payment and create a longer branch with generateblock so the transaction stays in the mempool instead of being re-mined. The balance should return to zero once the node reorgs.


rustaceanrob:
Nice, LGTM 36895c245759fcd372666d08dd1aaacec2c88e86

PR #94, opened 2026-08-13
test: Verify a reorg moves a re-mined payment to its new height

pzafonte:
Invalidate a block holding a silent payment and re-mine the transaction one block up on a longer branch. The balance should stay the same and the coin reports the new height after the node reorgs.

### DIFF
--- a/tests/wallet_reorg_remine.rs
+++ b/tests/wallet_reorg_remine.rs
@@ -1,0 +1,63 @@
+mod common;
+
+use std::time::Duration;
+
+use bitcoin::{Amount, BlockHash};
+use common::{fund_silent_payment, start_bitcoind, wait_for_core_mempool, TestNode};
+
+const SYNC_TIMEOUT: Duration = Duration::from_secs(60);
+const REORG_TIMEOUT: Duration = Duration::from_secs(60);
+const MEMPOOL_TIMEOUT: Duration = Duration::from_secs(30);
+const PAYMENT: Amount = Amount::from_sat(100_000_000);
+
+#[test]
+fn reorg_drops_and_restores_a_re_mined_payment() {
+    let core = start_bitcoind();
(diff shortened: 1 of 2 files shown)


tests/wallet_reorg_remine.rs
```diff
@@ -0,0 +1,61 @@
...
+        .parse::<BlockHash>()
+        .unwrap();
+    core.client.invalidate_block(funding_block).unwrap();
+    wait_for_core_mempool(&core, &funding_txid.to_string(), MEMPOOL_TIMEOUT);
+    println!("step 4/6: invalidated height {funding_height}, payment is back in the mempool");
+
+    let miner = core.client.new_address().unwrap().to_string();
+    core.client.generate_block(&miner, &[], true).unwrap();
+    core.client
+        .generate_block(&miner, &[funding_txid.to_string()], true)
```
rustaceanrob, line 53: We mine a block here that both creates a new most-work chain and confirms our transaction. I wonder if we can instead: mine the first block then mine another block _without_ the transaction relevant to our wallet that creates a new most-work chain. That should make our balance zero, and we can assert the balance is zero after that step. Next, we can mine a third block with the transaction and assert our balance is reset to `PAYMENT`.

This would allow us to avoid introducing string parsing for history, which can be inherently brittle. For instance we may want to change how history is printed in the future. That way the new `history` APIs in `test/common` can be omitted here

pzafonte: Done. Two blocks reorg the payment out so the balance goes to zero, then a third re-mines it and the balance returns to `PAYMENT`. Now uses `wait_for_balance_of` instead of parsing history, which polls the wallet's balance until it equals an exact expected `Amount`.

rustaceanrob (approved):
Tested locally and test logic looks good

PR #95, opened 2026-08-26
peer: Move the connection loop into a peer manager

pzafonte:
The loop runs inline in run(), and the connection writer lives in a shared slot so the stale block check can close it. A second connection would duplicate both. PeerManager owns a thread per connection and exposes the writers. Peer functions take `&NodeState` instead of `&mut NodeState`, since the threads share one. The manager still runs one connection, so behavior is unchanged.

### DIFF
--- a/src/bin/node.rs
+++ b/src/bin/node.rs
@@ -1,5 +1,5 @@
 use std::{
-    net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4},
+    net::{IpAddr, Ipv4Addr, SocketAddr},
     ops::DerefMut,
     sync::{
         atomic::{AtomicBool, Ordering},
@@ -27,7 +27,8 @@ use kernel_node::{
     ext::{ChainExt, DirnameExt, NetworkExt},
     ipc::IpcInterface,
     logging::Category,
-    peer::{BitcoinPeer, NodeState, TipState},
+    peer::{NodeState, TipState},
+    peer_manager::PeerManager,
     resolve_seeds,
     server_capnp::server,
     FatalShutdown, ScanEvent,
@@ -266,7 +267,7 @@ fn broadcast_transaction(
 fn run(
     network: Network,
     connect: Option<SocketAddr>,
-    mut node_state: NodeState,
+    node_state: NodeState,
     shutdown_rx: mpsc::Receiver<()>,
     addr_rx: mpsc::Receiver<Vec<AddrV2Message>>,
     block_rx: mpsc::Receiver<bitcoinkernel::Block>,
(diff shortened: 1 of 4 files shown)


rustaceanrob:
Lol, Github is falling apart recently. We should consider mirroring or moving this project to another git host since we only have a few devs active.

src/bin/node.rs
```diff
@@ -323,68 +324,20 @@ fn run(
...
-                    }
-                    break;
-                }
-            }
-        }
-        info!(target: Category::NODE, "Stopping net processing thread.");
-    });
+    let mut peer_manager = PeerManager::new(Arc::clone(&addrman), Arc::clone(&node_state), network);
+    if connect.is_some() {
+        peer_manager = peer_manager.max_peers(1);
```
rustaceanrob, line 342: If our `connect` peer is not live for whatever reason, maybe we should increase this? I don't have a strong opinion on how `connect` should behave though. We may also want to terminate the process if the peer we are expecting to be online is not. Doesn't have to be resolved here, but something to think about.

src/peer_manager.rs
```diff
@@ -0,0 +1,160 @@
...
+                    let selected = {
+                        let table = addrman.lock().unwrap();
+                        table.select().map(|record| record.network_addr())
+                    };
+                    let socket_addr = match selected {
+                        Some((AddrV2::Ipv4(ipv4), port)) => SocketAddr::from((ipv4, port)),
+                        Some((AddrV2::Ipv6(ipv6), port)) => SocketAddr::from((ipv6, port)),
+                        Some(_) => continue,
+                        None => {
+                            thread::sleep(Duration::from_secs(1));
```
rustaceanrob, line 114: Might want to eventually kill the process if we hit this condition enough times. That would indicate DNS is not working and the network is likely down or improperly configured.

rustaceanrob (commented):
Looks good, just a few minor comments. Running locally on mainnet and working as expected. I'll give you a chance to address review and see if a force push runs the CI successfully.

src/peer_manager.rs
rustaceanrob: Would be a use case for `FatalShutdown` IMO

src/bin/node.rs
pzafonte: I think we could let `--connect` take several addresses and set `max_peers` to however many were given.

rustaceanrob: Sure, seems fine. I think `connect` will mostly be used for faster IBD when we know a peer is co-located with us, so in practice I don't think this peer should timeout frequently, if at all.

src/peer_manager.rs
pzafonte: Done. Doing `FatalShutdown` after 60s.

pzafonte:
> Lol, Github is falling apart recently. We should consider mirroring or moving this project to another git host since we only have a few devs active.

I've seen various projects have discuss migrating away from GitHub. Unless there's an organization in the Bitcoin universe that opens up the option for projects like this to join their self-hosted solution, then I think GitLab is the next best option.

rustaceanrob (approved):
Looks good. We can probably do a similar breakout for address gossip and such. Synced to tip on mainnet with no issues.

yancyribbens:
> Unless there's an organization in the Bitcoin universe that opens up the option for projects like this to join their self-hosted solution,

there is that here: https://git.rust-bitcoin.org

Issue #96, opened 2026-08-26
Update `wallet` crate with `0.3.0` API 


rustaceanrob:
`0.3.0` is [out](https://docs.rs/bitcoinkernel/latest/bitcoinkernel/) and has some nice additions for us. Namely, for silent payments:

- [ ] Witness stack items
- [ ] Take a script as a slice reference

I think this should allow us to remove `bitcoin` from the wallet dependency, and also remove any conversions to/from the `bitcoin::Block`

rustaceanrob:
> I think this should allow us to remove bitcoin from the wallet dependency

I completely neglected signing here, so `bitcoin` will stay in `wallet`. In any case, #102 closes this.

PR #99, opened 2026-09-01
build: Update bitcoinkernel to 0.3.0

pzafonte:
process_block_header now returns a Result, and its success variant no longer carries a validation state, because the binding checks the mode itself. Match on Ok(ProcessBlockHeaderResult::Valid) and treat an error like a rejected header.


rustaceanrob (approved):
Nice. Going to leave #96 open until silent payments APIs are updated. Thanks.

PR #100, opened 2026-09-03
build: update silentpayments to 0.7.0

cygnet3:
Updates the silentpayments dependency from 0.5.0 to 0.7.0 (latest), introducing a few api changes:

1. functions that take an OutPoint as an argument now take a custom OutPoint struct instead of a (String, u32) pair.
2. renamed some getters, dropped the 'get_' prefix.
3. SilentPaymentAddress has been split into 2 separate structs, SilentPaymentCode and SilentPaymentKeyMaterial. The former represents a bech32-encoded address. The latter is the key material used to actually perform the calculations. This split was useful for PSBT's which provide only the key material, but not network info.

### DIFF
--- a/crates/wallet/Cargo.toml
+++ b/crates/wallet/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-silentpayments = { version = "0.5", features = ["receiving", "encode", "sending"] }
+silentpayments = { version = "0.7", features = ["receiving", "encode", "sending"] }
 bitcoin = { version = "0.32.8", features = ["rand-std"] }
 bitcoinkernel = "0.2"
 log = "0.4"
--- a/crates/wallet/Cargo.toml
+++ b/crates/wallet/Cargo.toml
@@ -4,7 +4,11 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-silentpayments = { version = "0.7", features = ["receiving", "encode", "sending"] }
+silentpayments = { version = "0.7", features = [
+  "receiving",
+  "encode",
+  "sending",
+] }
 bitcoin = { version = "0.32.8", features = ["rand-std"] }
 bitcoinkernel = "0.2"
 log = "0.4"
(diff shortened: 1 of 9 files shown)


crates/wallet/Cargo.toml
```diff
@@ -4,9 +4,13 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-silentpayments = { version = "0.5", features = ["receiving", "encode", "sending"] }
+silentpayments = { version = "0.7", features = [
```
yancyribbens, line 7: its better to keep fmt changes in a separate commit.  It looks like the only change here is from 0.5 to 0.7

cygnet3: Made a separate commit for that

crates/wallet/src/silentpayments/scanning.rs
```diff
@@ -123,8 +122,11 @@ pub(crate) fn scan_block_inner(
                 script_sig: btc_input.script_sig.as_bytes().to_vec(),
                 witness: btc_input.witness.iter().map(|item| item.to_vec()).collect(),
                 prevout_script: coin.output().script_pubkey().to_bytes(),
-                txid: btc_input.previous_output.txid.to_string(),
-                vout: btc_input.previous_output.vout,
+                outpoint: silentpayments::utils::OutPoint::from_txid_and_vout(
```
pzafonte: Might be worth constructing the outpoint like you do in sending, using `from_bytes` rather than `from_txid_and_vout`

tests/common/mod.rs
```diff
@@ -327,15 +312,19 @@ pub fn fund_silent_payment(core: &Node, sp_address: &str, amount: Amount) {
     };
     let prev_txout = coinbase.output[0].clone();
 
-    let sp = SilentPaymentAddress::try_from(sp_address).unwrap();
+    let sp = SilentPaymentCode::try_from(sp_address).unwrap();
     let partial_secret = calculate_partial_secret(
         &[(sender_sk, false)],
-        &[(prevout.txid.to_string(), prevout.vout)],
+        &[silentpayments::utils::OutPoint::from_txid_and_vout(
```
pzafonte: Same thing here.

crates/wallet/src/silentpayments/scanning.rs
cygnet3: Done, maybe I should add a `OutPoint::from_txid_bytes_and_vout` constructor to that outpoint struct?

pzafonte: I think that's a good idea.

cygnet3: Related PR: https://github.com/cygnet3/spdk/pull/167

PR #102, opened 2026-09-04
wallet: Avoid ser/de roundtrip of transactions in scanning

### LINKED ISSUE #101: Duplicate filtering of tr outputs
`scan_block_inner()` only proceeds if the output is a tr output [ref](https://github.com/kernel-node/kernel-node/blob/2214c95cef56523d408fd195813f6f3fa93dbedf/crates/wallet/src/silentpayments/scanning.rs#L108 ).

`scan_block_inner()` later calls `scan_transaction()` which once again filters out tr outputs [ref](https://github.com/kernel-node/kernel-node/blob/2214c95cef56523d408fd195813f6f3fa93dbedf/crates/wallet/src/silentpayments/scanning.rs#L50). since `scan_transaction()` only gets called if the output is tr, it doesn't appear that `scan_transaction()` also needs to filter fro tr outputs..

rustaceanrob:
We can use the scriptsig, witness elements, and take the `ScriptPubKey` type as a reference as of `kernel = 0.3.0`. I believe this also resolves #101


crates/wallet/src/silentpayments/scanning.rs
```diff
@@ -99,56 +81,70 @@ pub(crate) fn scan_block_inner(
...
+                .coin(idx)
                 .expect("input/spent-output count mismatch");
+            let outpoint = input.outpoint();
             let mut buf = [0u8; 36];
-            buf[..32].copy_from_slice(&btc_input.previous_output.txid.to_byte_array());
-            buf[32..].copy_from_slice(&btc_input.previous_output.vout.to_le_bytes());
+            buf[..32].copy_from_slice(&outpoint.txid().to_bytes());
+            buf[32..].copy_from_slice(&outpoint.index().to_le_bytes());
+            let witness: Vec<Vec<u8>> = input.witness_stack().items().collect();
+            let script_sig = input.script_sig().unwrap_or_default();
```
pzafonte: nit: I'm not sure if this _should be_ `expect` or `unwrap` or if it matters in this case.

pzafonte (commented):
ACK

crates/wallet/src/silentpayments/scanning.rs
rustaceanrob: Sure, I suppose we should not swallow this as the scriptSig can affect the silent payment tweak. I will change this to an `unwrap` in the off chance we get a kernel error and need to debug.

pzafonte (commented):
Tested, and seems to work as intended.

ACK 07eefbc

PR #103, opened 2026-09-04
node: Read block undo data on the scan thread

pzafonte:
On #50 the suggestion was to send the block over a channel and scan it on a consumer thread. I was under the impression that `get_block_tree_entry` returned `None` for just-connected blocks because get_block_tree_entry returned None for every hash in bitcoinkernel 0.2.0., so `read_spent_outputs` stayed in the block connected callback, and the callback reached the chainstate manager through a `OnceLock` because it is registered before the manager exists.

The lookup works in 0.3.0. and resolves the entry now. The callback sends the block and its hash, and the scan thread reads the undo data itself. The `OnceLock` is gone and the disk read is off the kernel's validation thread.

### DIFF
--- a/src/bin/node.rs
+++ b/src/bin/node.rs
@@ -18,9 +18,8 @@ use bitcoin::p2p::{
 use bitcoin::secp256k1::rand::random;
 use bitcoin::{hashes::Hash, BlockHash, Network, Transaction};
 use bitcoinkernel::{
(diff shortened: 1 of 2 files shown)


rustaceanrob (approved):
LGTM, tested and ran this locally for a few blocks. Thanks for the cleanup

PR #104, opened 2026-09-10
peer: Release buffered blocks the kernel can connect

pzafonte:
Buffered blocks are released only when their parent equals `TipState` which is a copy of the tip fed by `with_block_checked_validation`. That callback runs for every block that passes its checks, not only the one that becomes the tip so the copy can name a block that was stored without connecting, or one on a competing branch. However, while it does, the buffer drains only at the end of inventory flush.

Alternatively, we can ask the kernel to release a block once `get_block_tree_entry` finds its parent on the active chain. The parent no longer has to be the tip itself therefore `TipState` is no longer used anymore.

Originally, I had read the tip equality test as evidence that the kernel wanted blocks in chain order. It does not. Resolving a hash into a block tree entry is what lets the node ask where an arriving block's parent sits rather than comparing it against a hash the node already holds.


rustaceanrob (approved):
Cool, good to see some state-deduplication here.

PR #105, opened 2026-09-11
peer: Use ancestor lookups to build block locators

pzafonte:
A locator holds about 30 hashes. Currently the loop reaches each one by calling `prev()` a block at a time, making the work grow with the height of the chain. `BlockTreeEntry::ancestor` gets the height through the block index skip list, in O(log N).

Here the loop tracks a height instead of an entry, so it can use the i32 the kernel hands back. It also doubles the step after subtracting it, the way Core does, which moves the twelfth and later hashes one block closer to the tip.


rustaceanrob (approved):
LGTM

PR #106, opened 2026-09-17
Connect over Tor via Socks5 proxy

rustaceanrob:
We can connect to our peers over Tor using a socks5 proxy provided by the Tor daemon. This helps preserve user privacy at the cost of IBD time.

On mac, if I remember the steps correctly ,this can be tested with `brew install tor` and `brew services start tor` along with the configuration in the wizard or CLI.


src/lib.rs
pzafonte, line 78: Should the seed lookup here go through the proxy when one is set? As written it resolves over clearnet at startup, which seems like it'd leak that a node is starting even in Tor mode. For privacy, shouldn't  `--proxy` cover all outbound traffic?

src/peer.rs
```diff
@@ -494,15 +496,37 @@ impl BitcoinPeer {
             .set_service_requirement(ServiceFlags::NETWORK)
             .offer_services(ServiceFlags::WITNESS)
             .user_agent("/kernel-node:0.1.0/".into());
-        let (writer, reader, _) = conf.open_connection(socket_addr, TimeoutParams::new())?;
+        let port = destination.port;
+        let (writer, reader, _) = match proxy {
```
pzafonte: Could be simplified with a shared helper, that could be used in the feeler and broadcaster as well. 

```rust
fn connect(conf: ConnectionConfig, dest: &Destination, proxy: Option<&Socks5Proxy>, read_timeout: Duration) -> Result<(ConnectionWriter, ConnectionReader), p2p::net::Error>
```

src/peer_manager.rs
```diff
@@ -176,9 +227,212 @@ impl PeerManager {
...
+#[derive(Clone)]
+pub struct Broadcaster {
+    addrman: Arc<Mutex<AddrTable>>,
+    network: Network,
+    proxy: Option<Socks5Proxy>,
+}
+
+impl Broadcaster {
+    pub fn broadcast_transaction(&self, tx: &Transaction) -> bool {
+        let mut table = self.addrman.lock().unwrap();
```
pzafonte: Could we drop the `addrman` lock before the connect, so a slow dial doesn't block peer selection?

src/peer_manager.rs
```diff
@@ -176,9 +227,212 @@ impl PeerManager {
...
+                continue
+            }
+            Err(_) => return false,
+        }
+    }
+    false
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, std::hash::Hash)]
+pub struct Destination {
```
pzafonte, line 389: Would `Destination` fit better in its own module, since peer.rs and peer_manager import each other?

src/socks5.rs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct Socks5Proxy {
+    proxy: SocketAddr,
+    timeout: Duration,
+}
+
+impl Socks5Proxy {
+    pub const DEFAULT_TOR_PROXY: Self = Self {
```
pzafonte: `DEFAULT_TOR_PROXY` looks to be unused. Remove?

src/socks5.rs
```diff
@@ -0,0 +1,202 @@
...
+            }
+            ADDR_TYPE_IPV6 => {
+                // Read the IPv6 address and additional two bytes for the port
+                let mut buf = [0_u8; 18];
+                tcp_stream.read_exact(&mut buf)?;
+            }
+            ADDR_TYPE_DOMAIN => {
+                let mut len = [0_u8; 1];
+                tcp_stream.read_exact(&mut len)?;
+                let mut buf = vec![0_u8; u8::from_le_bytes(len) as usize];
```
pzafonte: Should the domain arm read the two port bytes too, like the IPv4/IPv6 arms do?

src/socks5.rs
```diff
@@ -0,0 +1,202 @@
...
+                tcp_stream.read_exact(&mut buf)?;
+            }
+            _ => return Err(std::io::Error::other("response failure")),
+        }
+        Ok(tcp_stream)
```
pzafonte, line 141: Doesn't this leave the 15s timeout set for the whole connection? On an idle peer the deadline would fire at 15s, before the 30s ping can prove it's still alive.

```suggestion
        tcp_stream.set_read_timeout(None)?;
        tcp_stream.set_write_timeout(None)?;
        Ok(tcp_stream)
```

pzafonte (commented):
Nice! I ran this against signet over a local Tor daemon and it syncs, and the onion encoding follows the Tor v3 spec. I recomputed the test's `.onion` from the pubkey and it matches.

All review questions are inline.

src/peer_manager.rs
rustaceanrob: Yes but I have some broader goals in separation of our modules so going to defer this one

src/lib.rs
rustaceanrob: I think we would have to roll our own DNS resolution to route over Socks5 because all this function accepts is the hostname we are looking for. I've tried doing this in the past and ran into issues at networks at places like coffee shops. Agree it would be better for privacy but I don't think it's possible without a heavy lift

src/socks5.rs
rustaceanrob: Good one

src/socks5.rs
rustaceanrob: Yeah I think so, good catch

src/peer.rs
rustaceanrob: Sure, seems like a good idea. Wasn't really sure where to put it in the commit sequence, so I amended the final one. I think this should go into a `net` module/crate in the future but for now it is in the peer manager.

src/peer_manager.rs
rustaceanrob: Yeah, also did that for `open_feeler`

